PR #54, opened 2015-01-20
Fix PHP fatal error.

colinmollenhour:


redox:
Replacing it with an `AlgoliaException` will make our API client think it's an application error (like 403, 404, etc...) -> in that case it will NOT retry on another host.

On the other hand, throwing an `Exception` will be catched and the API client will retry on the 2 other Algolia hosts -> maybe it's because the first host is not accessible or something.

So at the end, I think what we want is to replace `new Exception` by `new \Exception`; isn't it? What do you think @colinmollenhour?

colinmollenhour:
I hadn't looked into it that deeply so I think \Exception is best for the reasons you stated. Thanks, PR updated.

redox:
Thank you very much for your help @colinmollenhour !

PR #56, opened 2015-01-21
Improve lock contention with "Use Result Cache" option.

colinmollenhour:
Magento is designed to save the results of a query into the catalogsearch_result table, but these results get cleared so frequently and the use of this table creates so much additional lock contention and Algolia is so fast that this method really doesn't make sense. This pull request causes the default behavior to bypass the use of this table which actually makes loading the results page much faster and reduces database load and most importantly, lock contention with other important things like submitting orders.


redox:
Gonna review that tomorrow, but it looks really nice :) Thank you @colinmollenhour! :+1:

PR #60, opened 2015-02-06
Fix getProductJSON for listener's data and price type

gmoigneu:
Injected data by a listener to the algolia_product_index_before event were not used is the method.
The price should be sent as float to ensure Algolia engine is using it as a numerical value for ranges & sliders.


colinmollenhour:
Ahh, thanks for fixing this. It seems to be a result of a weird merge from @zainengineer's 9d0e2d2fc.

redox:
Thank you for contributing @gmoigneu !

PR #65, opened 2015-04-02
Support Magento installs that use prefixed MySQL tables

GLips:
Using `$this->_getWriteAdapter()` to grab a table name does not work for Magento installs that have prefixed MySQL tables (specified in `app/etc/local.xml`, under the node `config > global > resources > db > table_prefix`). To correctly retrieve the table's name, we need to go through the Magento DB singleton, which has more knowledge about Magento's internals and returns the correct, prefixed, table name.

All other uses of `$this->getWriteAdapter()` in the file seem to be fine since they aren't used to grab table names.


maxiloc:
@GLips Thanks for the contribution

GLips:
Gladly! Does this extension not quite support autocompleted search results yet?

PR #88, opened 2015-06-01
Fix image generation and calculation of custom ordered/stock_qty

SchumacherFM:
Allow bypassing getThumbnailUrl and getImageUrl because they trigger
the internal GD2 resizing functions. When you store your images in
e.g. S3 bucket you don't need that resizing functions and it even
generates a lot of errors due to missing files.

ordered_qty and stock_qty can be pre-calculated differently and
efficiently so the checks here can skip the default calculation in this
module.

Overall: your code style could be improved to match it more the Magento way.
Please check file permissions. There should be no file mode 755 on any
file or 777 on directories.

### DIFF
--- a/code/Helper/Data.php
+++ b/code/Helper/Data.php
@@ -574,52 +574,53 @@ class Algolia_Algoliasearch_Helper_Data extends Mage_Core_Helper_Abstract
 
         $customData['categories_without_path'] = $categories;
 
-        try
-        {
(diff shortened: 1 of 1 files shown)


maxiloc:
@SchumacherFM Thanks for your contribution

SchumacherFM:
Thanks a lot!

WesMage:
Thanks Cyrill! I didn't know you were interested in this product  @SchumacherFM

PR #91, opened 2015-06-08
changed is_int to is_numeric on pid job retry check

dlobb:
$pid is a string type here so is_int will never return true. I believe the original intention was to use is_numeric to ensure $pid was a number, regardless of data type.


maxiloc:
@dlobb Nice catch. Thanks

PR #92, opened 2015-06-08
changed how product filter gets products with visiblity search or catalog, search

dlobb:
Old product collection filtering method only retrieved products with visibility "catalog, search". Products with visibility "search" were not being included for whatever reason. Filtering the product collection this other way will instead get products with visibility "search" or "catalog, search".


redox:
:+1:

PR #105, opened 2015-06-16
Incrementing retry counter when job is retried instead of during exception

dlobb:
In current implementation, when cron process is killed due to timeout, no exception is thrown and thus the retry counter never gets incremented. This can result in an infinite loop since the job will continually get retried without the counter incrementing. Moving retry counter incrementing to when job is retried instead of during exception will account for this scenario.


maxiloc:
@dlobb Thanks. Fixed with 0ca0fac.

PR #106, opened 2015-06-16
Changing private to protected allowing proper magento rewrites

dlobb:
Changing the access modifiers to protected on methods and variables allows for proper Magento rewrites of these classes. In one case, we need to extend the getProductJSON to accommodate custom product data, but since this method relies on private variables/methods, we end up having to overwrite those as well. Having these private variables/methods be protected will allow for cleaner rewrites of these classes in case customization needs to be done.


maxiloc:
@dlobb Can you please do it on develop branch. A lot in those file has changed :). Thanks. Will add this info a Contributing.md

dlobb:
@maxiloc will do.

PR #111, opened 2015-06-17
Loading helper classes via magento helper loader

dlobb:
Looks like v1.2.0 is loading helper classes directly by name as opposed to using Magento's helper loader, preventing extension of the Algolia helper classes without some tedious modifying of several classes.

`$this->product_helper   = new Algolia_Algoliasearch_Helper_Entity_Producthelper();`
vs
`$this->product_helper   = Mage::helper('algoliasearch/entity_producthelper');`

Also, changed a few more private variables to protected.


maxiloc:
@dlobb I did not put them in the config file like Algolia_Algoliasearch_Helper. So does it still work

dlobb:
It should still work. Usually no need to add every helper class in the directory to that config file. Looks like it's still working on my local too.

maxiloc:
Ok I am gonna wait for tomorrow, to test it. And merge it then.  thanks

PR #161, opened 2015-07-29
feat: update algolia bundle to 1.4.0

vvo:


vvo:
cc @maxiloc

maxiloc:
@vvo thanks

PR #169, opened 2015-08-12
Fixed product attribute labels fetched from category instead of produ…

halk:
…ct entities

Let me know if it's the correct base branch.


maxiloc:
@halk Nice catch. Thanks for contributing

PR #195, opened 2015-09-11
Disabling the Mage_Review-module caused the indexing not to run

professorsloth:
Fixed a bug where having the Mage_Review-module disabled caused the indexing not to run.

Added a simple check for this. Runs without problems now.


redox:
Oh wow nice catch @professorsloth, what was the experienced behavior before that fix?

maxiloc:
:+1:

professorsloth:
@redox Got something similar to `no method load() in non-object` when it tried to load an object created with `Mage::getModel('review/review_summary')`. It crashed, stopping execution.

PR #200, opened 2015-09-18
Handling configurable, bundles and grouped products without options

kurtinge:
If the configurable, bundles or grouped products doesn't have any sub products. The $ids variable ends up beeing empty. This will result that the getProductCollectionQuery will return the whole product catalog as sub_products.


maxiloc:
@kurtinge Nice catch ! Thanks

PR #210, opened 2015-09-29
Do not do a query for categories if the product don't have any

kurtinge:
It seems like fetching a collection with an empty category array results in a a bad sql on at least Magento 1.9.0.1 
To avoid this we check if the product has any category ids attached to it and only doing the category collection if there is at least one category attached to the product.


maxiloc:
@kurtinge Thanks for contributing

Issue #185, opened 2015-08-28
On install - disable everything until index is complete


sam327:
Apologies if this is long...  

When you install the algolia search and put in your various api keys, it instantly takes over the search results. The problem is that none of the indexes have been built yet. This means that for a little bit of time, we have blank search results. 

So I suggest adding a toggle to turn off the search results - so while I'm getting the plugin started, my site continues to use the regular magento catalog search while I get algolia configured and the indexing done. 

Then I can switch on the components manually when the index is ready.

My preferred setup configuration would be something like: 
1. install plugin 
2. add license / api keys
3. configure search options 
4. configure indexing
5. run indexing (all at once or via cron)
6. enable search results
7. enable other features as desired

Didn't really notice while doing development, as I'd installed, then spent a bunch of time doing design stuff... But when I rolled it up to the production environment, I had about 20 minutes of blank search results.

maxiloc:
@sam327 Yes that is link to the #175 issue. To do that we need to find a way to disable Algolia. The thing is the extension goes deep into magento indexing and event system and I didn't yet find a way to handle that correctly.

maxiloc:
We will work on this one for the next release. We have some path to explore. cc @Floddy

Floddy:
Dunno if you've found a way yet, but the solution I've seen the most is:
1. config option, enble, yes/no
2. a helper function that responds true/false if enable is yes, and api key isn't empty
3. for every event observer, start with something like:
public funtion meepevent(Varien_Event_Observer $observer)
{
if(!$helper->isEnabled()) {
    return;
}
4. for blocks, etc, an override of _toHtml:
protected function _toHtml()
{
    if(!$helper->isEnabled()) {
        return '';
    }
    return parent::_toHtml;
}
5. for controllers, pretty much:
public function myAction()
{
    if(!$helper->isEnabled()) {
        parent::myAction();
        return;
    }
}

etc etc. :)

maxiloc:
@Floddy Thanks will have a look at that

maxiloc:
Implemented with 3a87884.

@sam327 Everything will be enabled by default but can easily be disabled afterwards.
You will then be able to do have the exact workflow you described

maxiloc:
Will be available with next release

PR #215, opened 2015-10-02
Added feature to enable/disable alogoliasearch module

manish172:


maxiloc:
@manish172 Thanks that is a good start to fix #185

I can not merge that as it is right now. The main reason is that we want to be able to index to Algolia launching full reindex even if store is disable. Not sure yet of the best way to implement that. Maybe we need two configs for disableling completly.

maxiloc:
@manish172 I just implemented it with 3a87884. Your work was the foundation but I needed to add a lot of logic to handle #185.

manish172:
Great! Thanks for adding this. It's a big addition for the customers who has big catalog and many stores which they are using only in backend for ebay/amazon etc. 

There is one more thing related to indexing performance. You had included ordered_qty and rating_summary in function rebuildStoreProductIndexPage (line 371 and 372) but if there there thousands of order then it makes indexing very slow. 

If we have some admin settings to include/exclude ordered_qty and rating_summary then it would be easy to exclude these data from indexing for stores which have thousand of products.

Please let me know your feedback on above mentioned indexing performance issue.

maxiloc:
@manish172 The thing is having ordered_qty remove from the customRanking will hurt relevance really bad. We are aware of the performance issue, but for now I think it's better to have a long indexing and good relevance than the opposite.
I am open to suggestions on how to optimise those queries but removing them from indexing is not a good solution.

manish172:
@maxiloc Could you please let me know that when are you  going to merge your recent changes into master branch?

maxiloc:
@manish172 Do not know yet. I still need to do some testing

manish172:
@maxiloc ok. no problem. please let me once you will merge your changes.

PR #224, opened 2015-10-13
Upload prices with taxes again

rbrown:
If Magento is set to display both prices, the current code will only export the
price without tax. This patch rewrites the handlePrice method to include prices with and without tax, when displayBoth is set.

Is this or something like it likely to be accepted? I realise this patch looks really horrible, but I refactored the handlePrice method at the same time because it was quite long and there would be a lot of duplication.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -233,122 +233,173 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         }
     }
 
-    private function handlePrice(&$product, $sub_products, &$customData)
+    private function getPrice(&$product, $price, $returnWithTax) {
+        return (double) Mage::helper('tax')->getPrice(
+            $product,
+            $price,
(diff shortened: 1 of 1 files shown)


maxiloc:
@rbrown Yes that is very cool and definitely something we want to handle.
Several points before I can merge this:
- Can you let the handling of prices in one function, it's ok to have helper functions getPrice, getFields,  but I would rather have initHandlePrice, handleGroupPrices, ... inside handlePrice. I think there will not be to much code duplication
- Can also modify the topsearch.phtml file to handle the display correctly

rbrown:
So, I've pushed a new version, this one should actually work for the three different tax display settings. I haven't updated the template because I don't think it's ever displayed both? I looked in version 1.4.4 and there if the with_tax prices were available it used them, otherwise it used without, and I don't want to change the page layout too much.

maxiloc:
@rbrown I made some small modifications and fix small issues but it's was working pretty well.
Thanks for contributing

PR #226, opened 2015-10-15
Settings for image size and image type

aude:
So, at my company we have had problems with indexing requiring just too much resources, and end up in undefined behavior. One of those have been placeholder images being indexed.

The proposed functionality would be to have a setting for image size. As many Magento stores already have a lot of images in a fixed size in the cache, eg. 265x265, it would be beneficial to let Algolia Search for Magento use those images as well, and thus maybe cut the resource use.

There is also a setting added for which product image to use: Base Image, Small Image or Thumbnail.


maxiloc:
@aude It seems your forgot Algolia/Algoliasearch/Model/System/Imagetype.php in the commit

aude:
@maxiloc You are very right. Creds for checking it!

maxiloc:
I merged it in the develop branch. Thanks for contributing.
It will be available with next release

PR #229, opened 2015-10-16
Making products mappable to standard Magento types

JosephMaxwell:
We have developed several custom product types for a client. These product types are built on Magento's standard product types and add extra functionality (but don't change how the original product works).

As it is, Algolia doesn't recognize these. I made a simple mapping configuration to be able to map custom product types to Magento product types.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -6,9 +6,25 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
 
     protected static $_predefinedProductAttributes = array('name', 'url_key', 'description', 'image', 'thumbnail');
 
+    protected $_productTypeMap;
+
     protected function getIndexNameSuffix()
     {
         return '_products';
+    }
+
+    protected function getMappedProductType($originalType)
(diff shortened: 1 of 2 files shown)


maxiloc:
@JosephMaxwell Looks good. Can you target the develop branch instead of the master one ?

code/Helper/Entity/Producthelper.php
```diff
@@ -6,11 +6,27 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
...
+
     protected function getIndexNameSuffix()
     {
         return '_products';
     }
 
+    protected function getMappedProductType($originalType)
+    {
+        if (!isset($this->_productTypeMap[$originalType])) {
+            $mappedType = (string)Mage::app()->getConfig()->getNode('default/algoliasearch/product_map/' . $originalType);
```
maxiloc: Can you encapsulate this in https://github.com/algolia/algoliasearch-magento/blob/develop/code/Helper/Config.php

maxiloc: You can then access the function via $this->config->your_function

PR #230, opened 2015-10-16
Making product types mappable via configuration

JosephMaxwell:
Thanks for the comments. I have moved my work to the develop branch and the lookup to the config helper.


JosephMaxwell:
Do you have a rough idea as to when you can merge these changes? We are looking to push some code that uses these updates.

maxiloc:
Now :smile:. It will be part of the next release, probably early next week

maxiloc:
Thank for contributing

JosephMaxwell:
Awesome, thank you.

Issue #228, opened 2015-10-16
AvS_FastSimpleImport and catalogsearch_resource


rbrown:
If I import products using [AvS_FastSimpleImport](https://github.com/avstudnitz/AvS_FastSimpleImport) it fires of an [event](https://github.com/avstudnitz/AvS_FastSimpleImport/blob/master/src/app/code/community/AvS/FastSimpleImport/Model/Import/Entity/Product.php#L538) to rebuild the catalogsearch index for the affected products. It looks like this extension is [deliberately doing nothing](https://github.com/algolia/algoliasearch-magento/blob/develop/code/Model/Resource/Fulltext.php#L53) to avoid index twice, but because AvS_FastSimpleImport has used ImportExport, no other events that algolia is watching have been fired. Is there an obvious way around that?

maxiloc:
@rbrown Yes I understand the issue. If we remove this condition everytime you update a product both Algolia indexer and catalog search will index, which is not good.
What I would do is listen for "fastsimpleimport_reindex_products_before_fulltext" event and then run Algolia indexer from it
Does that makes sense for you ?

rbrown:
Sort of. And the call the Observer->rebuildProductIndex. Does this miss the point where it checks to see fi a product has parents?

maxiloc:
Yes you will need to add this logic https://github.com/algolia/algoliasearch-magento/blob/master/code/Model/Indexer/Algolia.php#L313-L321 in the event

rbrown:
Great, thanks

PR #231, opened 2015-10-19
Reindex after AvS_FastSimpleImport product import

rbrown:
Is this the sort of thing you'd be willing to accept for #228 or would you prefer if I just kept this in my own module?

After AvSFastSimpleImport updates products, this will:
* Reindex the affected products.
* Reindex the current categories of the affected products.

It won’t:
* Reindex any categories that affected products were removed from.
* Remove products that are deleted. (it will remove disabled/not visible
in search)
* Do anything with category imports or category/product imports.

### DIFF
--- a/code/Model/Observer.php
+++ b/code/Model/Observer.php
@@ -195,4 +195,55 @@ class Algolia_Algoliasearch_Model_Observer
 
         return $this;
     }
+
+    public function rebuildAfterFSIProductImport($event) {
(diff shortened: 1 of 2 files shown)


maxiloc:
@rbrown We discussed it with the team today, having extensions specific code/config is not something we want right now. So yes for now you should keep it in your own module.

rbrown:
No problem, thanks

PR #232, opened 2015-10-19
Code improvement and adding total sales for product ranking

JosephMaxwell:
In troubleshooting a problem with images, I saw that Algolia is overriding the `catalog/image` helper. Instead of totally overriding it, we can just subclass it. This will prevent other module overrides from interfering with Algolia.

I also added the total amount sold to the indexer. Our client has a case that they may sell a large quantity of a product, but low dollars, but they want to promote the products that have more dollars sold.


maxiloc:
@JosephMaxwell Thanks. Very good. I just change put back the default config as it does not make sense to have both ordered_qty and total_ordered at the same time for the default config

JosephMaxwell:
Thank you for your quick assistance! We appreciate it.

redox:
Thanks for contributing @JosephMaxwell, much appreciated!

PR #234, opened 2015-10-20
Changed image resize to avoid rewrite of catalog/image helper

kurtinge:
Instead of doing a rewrite of catalog/image just use the algoliasearch/image helper directly in the code.

Doing rewrites will often cause conflicts with other modules so avoiding it is preferred.


maxiloc:
@kurtinge Thanks. It's already done on the develop branch.

kurtinge:
Ah. Didn't check that.

PR #235, opened 2015-10-23
Adding mechanism to exclude specific products

JosephMaxwell:
The `exclude_from_search` attribute is now added so any product can be excluded from Algolia's search results.

Please disregard the changes to `topsearch.phtml`. I was having trouble working with my repo that (they are just spacing/tab formatting changes).


maxiloc:
@JosephMaxwell Actually you can remove the search from visibility to exclude it from search. 
image

Adam told me about what you are trying to do but didn't think about that directly.

Tell me if that would work for you

JosephMaxwell:
Hmm. I didn't think about that. That makes good sense, and that should work great.

PR #237, opened 2015-10-27
Only add categories under the current root category

kurtinge:
This will only index the categories under the current category for each store.

If you have several category trees for each store, and have the same products in multiple category trees this patch will prevent that all categories trees are indexed for each store. This will only index the categories that derive from the root category for the current store.


maxiloc:
@kurtinge this is nice,  Thanks for contributing.

PR #240, opened 2015-10-29
Add tax_class_id to product collection to be able to calculate price with tax

kurtinge:
Add tax_class_id to product collection to be able to calculate price with tax. This is useful if you have your prices excluding tax in your catalog but are showing the prices with tax in frontend.


maxiloc:
@kurtinge I do not have any issue with price computation. When I change to:
image
image
prices are updated correctly when reindexing.
Can you give me more details on how to reproduce the issue.

kurtinge:
The issue was on a customer installation running EE 1.14.2.1, but I'm not able to reproduce it on my own CE 1.9.2.2 dev install. I'll do some more digging.

PR #241, opened 2015-10-29
Simple cleanup

aude:


maxiloc:
@aude Nice catch. Thanks for contributing

PR #243, opened 2015-11-03
Set configurable product out of stock, if all the children is out of stock

kurtinge:
Sets the configurable product as out-of-stock when all the children is out of stock.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -646,6 +646,11 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
                         {
                             $customData[$attribute['attribute']] = array_values(array_unique($values));
                         }
+                        else {
+                            // Set main product out of stock if all
+                            // sub-products is out of stock.
+                            $customData['in_stock'] = 0;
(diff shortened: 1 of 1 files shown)


redox:
Ah good catch @kurtinge; I'll let @maxiloc handle the merge but it looks good :+1:

maxiloc:
@kurtinge I am not sure why we need this, Magento is supposed to put it out of stock by himself, no ?

Also the modification you made does not do exactly what it supposed to. Right now, it put the main product out of stock if one of the additional attribute is not set in all the children even if they are in stock.

maxiloc:
@kurtinge I am closing this, feel free to re-open if needed

kurtinge:
The Configurable product in Magento (or even grouped products) does not update the stockitem  automaticly when all the children is out of stock.

The check on line 588 will always be true as long as the status on the configurable product is set to "in stock":
$customData['in_stock'] = (int) $stockItem->getIsInStock();

Since Algolia is using this to see if the product is sellable or not, the configurable product should be set out of stock if all the sub products is out of stock.

kurtinge:
I've made a new pull request for this, fixing the issue with the value check.

PR #246, opened 2015-11-12
German locale has been added

olegpuzanov:


maxiloc:
@olegpuzanov I am not sure this should be part of the extension. If we go that way we gonna have a lot of language to maintain and I am not sure this is something we want. I will discuss that with the team and come back to you.

maxiloc:
@olegpuzanov Sorry for the late response. This is something we could imagine putting in the extension. We are currently working on a new major release so if you want to update your PR after the release we can definitely merge it.

maxiloc:
@olegpuzanov I am closing this, If you want to update it with the new release and submit a new PR, We would be happy to merge it.

PR #253, opened 2015-11-17
Handle MSRP price

Zifius:
Need to check for the case when both special price and MSRP are set, don't merge yet I'll test and update if needed

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -326,6 +326,13 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
                     $customData[$field]['default'] = $special_price;
                     $customData[$field]['default_formated'] = $product->getStore()->formatPrice($special_price, false);
                 }
+
+                // Handle MSRP price, the attribute should be indexable
+                $msrpPrice = $product->getMsrp();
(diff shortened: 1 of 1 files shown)


maxiloc:
@Zifius For me MSRP price should not be a special price, it should replace the default price, no ? Do we agree than if map if enabled it should not display the default price but the manufacturer price ?

Zifius:
1. Well I think MSRP > normal price > special price. Also according to https://mage-user-guide.atlassian.net/wiki/display/MUG/9.6+Minimum+Advertised+Price if special price set it should "considered to be the actual price for MAP" which I don't really get
2. Agreed, according to the page above "will not be indexed by search engines", thought in the project I work on MAP setting is not enabled and MSRP used as "old" (and bigger) price

maxiloc:
@Zifius With this commit https://github.com/algolia/algoliasearch-magento/commit/d82d8634758bf0ecf0776cb155fd78073d50ec3f the behavior is that if you have map enabled it will just not index price.
Are enabling map and want price to show up ?

Zifius:
No, not enabling the MAP. Just want MSRP to be indexed as "old" price to "default_original_formated" field

maxiloc:
Yes so I guess this is not something we want in the extension because Magento is not handling that this way. But yes what you do is definitely good for what you are trying to do so you keep the fix on your instance

Zifius:
Agreed!

PR #267, opened 2015-12-08
Set configurable product out of stock, if all the children is out of stock.

kurtinge:
The Configurable product in Magento (or even grouped products) does not update the stockitem  automaticly when all the children is out of stock.

The check on line 588 will always be true as long as the status on the configurable product is set to "in stock":
$customData['in_stock'] = (int) $stockItem->getIsInStock();

Since Algolia is using this to see if the product is sellable or not, the configurable product should be set out of stock if all the sub products is out of stock.


maxiloc:
@kurtinge Yes your were right.
Thanks for this PR, much appreciated.

PR #268, opened 2015-12-09
Hotfix for special characters in json

sergy-gzh:
Use case:
In dutch language 
```Pages -> Pagina's```
That is breaking json and whole top search block.
Added json encoding to avoid it.


maxiloc:
@sergy-gzh Thanks for contributing, Can you make the pull request on the develop branch instead ?

sergy-gzh:
@maxiloc In develop branch this file is empty. Probably because whole frontend part is refactored.
Not sure if this bug can be reproduced on develop. Will test tomorrow.

maxiloc:
Yes your are right there is indeed a lot of changes and those title are directly in the admin panel, sorry about that. I will make sure this issue is fiedx for the develop branch directly. Just created an issue to keep track of it (#209)

PR #276, opened 2015-12-16
Changed code to use Magento method calls instead of hard coded values

lazyguru:
See comments on code for fix for #273. This uses Magento method calls instead of relying on hard-coded values which could change (or expected behavior of values could change. Using the method calls should have Algolia behave the way the user would expect it to based on native Magento behaviors)


maxiloc:
@lazyguru Thanks for contributing, much appreciated.

Issue #277, opened 2015-12-21
Product URLs are broken


lazyguru:
The change in this commit d2ed177b6cdf6515017214cd38fd01dba1836e29 breaks product URLs.  The code should instead be calling $product->getProductUrl()

PR #278, opened 2015-12-21
Fix for product URLs

lazyguru:
Fix for #277.  NOTE: This should be done as a hotifx release since the current code will make it so that all products generate a 404 from search or categories when using Algolia.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -446,7 +446,7 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         $customData = array(
             'objectID'          => $product->getId(),
             'name'              => $product->getName(),
-            'url'               => Mage::getBaseUrl() . $product->getRequestPath(),
+            'url'               => $product->getProductUrl(),
             'visibility_search'  => (int) (in_array($visibility, $visibleInSearch)),
             'visibility_catalog' => (int) (in_array($visibility, $visibleInCatalog))
         );


maxiloc:
@lazyguru We changed that, because it was failing for some customer. It works fine on https://magento.algolia.com
Do you know why this method is failing for you ?

lazyguru:
It is failing, because that is not the right way to get a product URL.  It returns the stub and not the actual URL that Magento is listening for (eg it doesn't have the ".html" on the end)

maxiloc:
@lazyguru Thanks

maxiloc:
@lazyguru 
$product->getProductUrl() does not seems to take into account the store view url but only the website url.
Do you have an idea on how to fix this ?

lazyguru:
It would need to be tested, but you might be able to use $product->getUrlInStore() instead.  I think the issue might be that the indexer is running as the admin store instead of running for each store (no way to confirm this on my end as we only have a single website/store being used)

maxiloc:
Thanks,
Yes this is definitely the issue, emulation that we are using here https://github.com/algolia/algoliasearch-magento/blob/master/code/Helper/Data.php#L514-L531 should do the job but is not giving the expected result. I already tried getUrlInStore but will give it another try next week.

Zifius:
That's what I was fixing in the commit that got reverted, not sure thought why it breaks single stores, perhaps the url suffix is disabled? Anyhow would be nice to have a solution that work for all

maxiloc:
I tried it again, seems to work fine (don't know what happened last week). @Zifius Are you running on the last version ? Can you give me a way to reproduce it ?

Zifius:
On latest stable 1.5.2. It should be easy to reproduce as setting different store url on store views grouped under single store. No time today but I will try tomorrow with a clean Magento too

PR #280, opened 2015-12-22
tweaked wording on front end

sammarcus:


maxiloc:
@sammarcus Thanks for contributing

PR #285, opened 2015-12-31
Fix missing group_id variable

lazyguru:
$group_id is not set anywhere so it throws an error in the template file when customer groups are enabled


maxiloc:
@lazyguru Thanks

PR #287, opened 2016-01-08
Use protected instead of private for methods and fields

lazyguru:
Be nice to developers that need to extend your code.  Use protected instead of private


lazyguru:
Sorry, I targeted the wrong branch.  Let me know if you would like me to reissue this against develop instead

maxiloc:
@lazyguru Yes can you target develop please

PR #289, opened 2016-01-08
Use protected instead of private for methods and fields

lazyguru:
See https://github.com/algolia/algoliasearch-magento/pull/287 targeting develop branch this time instead of master


maxiloc:
@lazyguru Thanks

PR #294, opened 2016-01-15
Add config setting to control logging

lazyguru:
Added setting to allow disabling logging (and disable it by default) to avoid creating a large algolia.log file.  This is needed as the main logging setting might be enabled but merchant will not want (or need) this log enabled


maxiloc:
@lazyguru Can you use this class https://github.com/algolia/algoliasearch-magento/blob/master/code/Helper/Config.php to handle the config getter ?

lazyguru:
Done

maxiloc:
Thanks

PR #315, opened 2016-02-10
Correct path to placeholder image

kurtinge:
The placeholder is located in skin and in the latest version of the module it's assumed
that all pictures are located in /media
This makes the path to placeholder images wrong.
This fix is hacky, but should work

Feel free to reject this in advantage of a better fix.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -579,6 +579,7 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         $customData['categories'] = $categories_hierarchical;
 
         $customData['categories_without_path'] = $categories;
+        $baseSkinDir = Mage::getBaseDir('skin').DS;
 
         if (false === isset($defaultData['thumbnail_url']))
         {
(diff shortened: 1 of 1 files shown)


lazyguru:
The other way to deal with this is to rip out a bunch of code from the Image helper class:

```
<?php

class Algolia_Algoliasearch_Helper_Image extends Mage_Catalog_Helper_Image
{
    /*
     * Subclass to be able to catch the error
     */
    public function toString()
    {
        return $this->getUrl($this->__toString());
    }

    public function getUrl($imageUrl)
    {
        $baseDir = Mage::getBaseDir('media');
        $path = str_replace($baseDir . DS, "", $imageUrl);
        $path = str_replace(array('https://', 'http://'), '//',$path);
        $url = str_replace(DS, '/', $path);

        return $url;
    }
}

```

I think the original intent was to strip out the http/https so that image URLs would load correctly regardless of whether you were viewing the page over SSL or not.  My version above allows the native Magento function to generate the proper image URL but then removes the http/https part

maxiloc:
@kurtinge Thanks. I think @lazyguru version is cleaner. I will try that tomorrow

JanPetr:
Hey @kurtinge,
I went through your PR and I'm not sure if I got your problem.

In original code we already use `baseSkinUrl` and it's value depends on filled value in `Configuration > Web > (Un)Secure`. I think it's not good idea to _hardcode_ `../skin` into extesion's code as it would break other Magento instances.

Could you please provide some more information or confirm that it's OK? I'm really not sure I got it right. Thanks ;)

JanPetr:
So in the end I got your point and the issue should be fixed in this PR #383. So I'm closing this PR. If you have any doubts or questions feel free to leave a message.

PR #329, opened 2016-03-07
Excluded attributes array now separate function

thaiphan:
This change allows us to better customise which attributes we want to exclude.


maxiloc:
@thaiphan I'll be happy to merge this. Just re-create it on the develop branch

PR #339, opened 2016-03-17
Make child and parent skus searchable

rbrown:
Currently Producthelper->getObject only returns attribute values from child
products if the parent product attribute value is null. We would like to be able to
search for the sku of both parent and child products. This patch adds an
exception to getObject to include all the skus in the returned object.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -474,6 +474,15 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         }
     }
 
+    protected function getValueOrValueText(Mage_Catalog_Model_Product $product, $name, $resource)
+    {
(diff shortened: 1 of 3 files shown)


rbrown:
I'm not sure this is the best way to attack this, I can think of some other attributes where this might be useful, but sku is the one that was annoying us

code/Helper/Entity/Producthelper.php
```diff
@@ -116,6 +116,7 @@ public function setSettings($storeId)
 
         foreach ($this->config->getProductAdditionalAttributes($storeId) as $attribute)
         {
+            $attribute_name = $attribute['attribute'];
```
maxiloc: Do we need this line ?

maxiloc:
Except for the small comment, looks good. Can you also rebase your commits ?

rbrown:
Apparently I can't rebase my commits properly, I'll send a new pull request.

PR #340, opened 2016-03-17
Skus

rbrown:
OK, here it is. I got rid of the excess line, it was a holdover from an overzealous search and replace.


maxiloc:
Thanks for contributing

PR #347, opened 2016-04-04
Added cache for popular queries suggestions.

dannydnz:
We found that the getPopularQueries method was quite slow. The query used by this method was taking ~ 0.200ms. Instead of creating an index (ideal solution) for this we decided that a quick win was to save the popular query data to cache.

Let me know your thoughts / improvements.

### DIFF
--- a/code/Helper/Entity/Suggestionhelper.php
+++ b/code/Helper/Entity/Suggestionhelper.php
@@ -68,8 +68,8 @@ class Algolia_Algoliasearch_Helper_Entity_Suggestionhelper extends Algolia_Algol
                     $tags = array(
                         Mage_CatalogSearch_Model_Query::CACHE_TAG
                     );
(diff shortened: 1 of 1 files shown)


maxiloc:
@dannydnz Looks good, the only thing is I am not sure we want to have a lifetime cache on this query.

dannydnz:
@maxiloc Hmm it is optional but you can't guarantee that people are refreshing their cache as often as the popular queries change. It might make sense to set this to a day or longer rather than using the store cache settings. Whats your thoughts ? :smile:

maxiloc:
I think we can go for a week. Can you make the change ?

dannydnz:
Sure. Sorry it took so long. :smile:

maxiloc:
I think we're good. Can you re-open it targeting the develop branch instead ?

PR #350, opened 2016-04-07
store_id value in algolia_additional_section_item_index_before event

Pihtt:
Current store_id var seems to be usefull in this event.


maxiloc:
@Pihtt Look's good. Can you re-open it targeting develop instead of master, I'll be happy to merge it then.

Pihtt:
Done

PR #352, opened 2016-04-07
Add cache for popular queries

dannydnz:


maxiloc:
@dannydnz Thanks for contributing

PR #353, opened 2016-04-07
multiselect array unique

Pihtt:
getColumnValues() for multiselect attributes returns ids imploded by comma.
Each row should be exploded. If not,final array contains much duplication.


maxiloc:
Nice catch,

is
```
array_unique(explode(',', implode(',', $products->getColumnValues($attributeCode)));
```
not doing the same ?

Pihtt:
Yes, doing the same..
array_keys(array_flip()) is just faster than array_unique()
http://php.net/manual/fr/function.array-unique.php#70786

maxiloc:
Ok works for me

maxiloc:
Thanks for contributing :+1:

PR #355, opened 2016-04-14
Remove products from Algolia index if disabled

thaiphan:
This change updates the re-indexing functionality to also remove products that should no longer show up in Algolia.

### DIFF
--- a/code/Helper/Data.php
+++ b/code/Helper/Data.php
@@ -536,6 +536,63 @@ class Algolia_Algoliasearch_Helper_Data extends Mage_Core_Helper_Abstract
         $this->logger->stop('rebuildStoreProductIndexPage '.$this->logger->getStoreName($storeId).' page '.$page.' pageSize '.$pageSize);
     }
 
+    /**
+     * Removes products that are no longer indexable, i.e. they are disabled,
+     * from the specified Algolia index. You can provide an array from product
+     * IDs that you want to query. Otherwise, the entire product collection will
+     * be queried.
+     *
+     * @param string $storeId
+     * @param array $productIds
+     */
+    public function removeNonIndexableProducts($storeId, array $productIds = array())
+    {
+        $this->logger->start('REMOVE FROM ALGOLIA');
+
+        $nonIndexableProductIds = $this->getNonIndexableProductIds($storeId, $productIds);
+        if (count($nonIndexableProductIds) > 0) {
+            $this->removeProducts($nonIndexableProductIds, $storeId);
+        }
+
+        $this->logger->stop('REMOVE FROM ALGOLIA');
+    }
+
+    /**
+     * Returns an array of the IDs of products that are no longer indexable,
+     * i.e. they are disabled. You can provide an array from product IDs that
+     * you want to query. Otherwise, the entire product collection will be
+     * queried.
+     *
+     * @param string $storeId
+     * @param array $productIds
+     * @return array
+     */
+    public function getNonIndexableProductIds($storeId, array $productIds = array())
+    {
+        $nonIndexableProductIds = array();
+
+        /** @var Mage_Catalog_Model_Resource_Product_Collection $products */
+        $products = Mage::getModel('catalog/product')->getCollection();
+        $products
+            ->addStoreFilter($storeId)
+            ->addAttributeToFilter('type_id', 'configurable')
+            ->addAttributeToFilter('status', Mage_Catalog_Model_Product_Status::STATUS_DISABLED)
+        ;
+
+        if (!empty($productIds)) {
+            $products->addAttributeToFilter('entity_id', array(
+                'in' => $productIds
+            ));
+        }
+
+        foreach ($products as $product) {
+            /** @var Mage_Catalog_Model_Product $product */
+            $nonIndexableProductIds[] = $product->getId();
+        }
+
+        return $nonIndexableProductIds;
+    }
+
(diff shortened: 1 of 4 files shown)


code/Helper/Data.php
```diff
@@ -511,6 +512,65 @@ public function rebuildStoreProductIndexPage($storeId, $collectionDefault, $page
...
+     */
+    public function getNonIndexableProductIds($storeId, array $productIds = array())
+    {
+        $nonIndexableProductIds = array();
+
+        /** @var Mage_Catalog_Model_Resource_Product_Collection $products */
+        $products = Mage::getModel('catalog/product')->getCollection();
+        $products
+            ->addStoreFilter($storeId)
+            ->addAttributeToFilter('type_id', 'configurable')
```
lazyguru: Is it intentional that you are filtering for only configurable products? This conflicts with the PHPDOC that claims it is the entire product catalog

code/Helper/Data.php
```diff
@@ -511,6 +512,65 @@ public function rebuildStoreProductIndexPage($storeId, $collectionDefault, $page
...
+     * be queried.
+     *
+     * @param string $storeId
+     * @param array $productIds
+     */
+    public function removeNonIndexableProducts($storeId, array $productIds = array())
+    {
+        $this->logger->start('REMOVE FROM ALGOLIA');
+
+        if (!$this->config->getShowOutOfStock()) {
```
lazyguru: What if I want to show out of stock but I still want disabled products removed?  Maybe use a different flag?

lazyguru:
I'm +1 for this functionality, just a couple of thoughts on your code though (see inline comments)

code/Helper/Data.php
maxiloc: Yes I would rather add this condition to the collection creation.

A bit like this => https://github.com/algolia/algoliasearch-magento/blob/master/code/Helper/Entity/Producthelper.php#L82-L83

code/Model/Observer.php
```diff
@@ -203,14 +207,18 @@ public function rebuildProductIndex(Varien_Object $event)
...
         }
         else
         {
-            if (! empty($page) && ! empty($pageSize))
+            if (! empty($page) && ! empty($pageSize)) {
+                $this->helper->removeNonIndexableProducts($storeId, $productIds);
                 $this->helper->rebuildStoreProductIndexPage($storeId, $this->product_helper->getProductCollectionQuery($storeId, $productIds), $page, $pageSize);
-            else
+            } else {
+                $this->helper->removeNonIndexableProducts($storeId, $productIds);
```
maxiloc: You can probably call it only once before the "if"

maxiloc:
@thaiphan 
Looks good. I also put some comments. Tell us what you think.

Also just to be sure, when is this function useful ? Because in most of the cases, when you disable a product I guees it should get remove from algolia already

code/Helper/Data.php
thaiphan: Uh yeah, fair point. Algolia is pretty expensive so I didn't want to process all of the simples since that's like 30,000 requests for us. I guess I can just disable products that are not visible online!

code/Helper/Data.php
thaiphan: Yeah I will address this.

thoaionline:
This is useful when products are disabled in bulk (indexing not triggered). See also #357.

thaiphan:
Hey guys, I'm still here. Just been busy.

maxiloc:
@thaiphan do you think fixing #357 won't be enough ?

thaiphan:
Hey @maxiloc, yeah my change should address that issue. However, I haven't made the necessary amendments yet! Please hold off on merge this in until then!

maxiloc:
Yes the thing is I would rather fix the hook, than adding a logic to "patch the issue" but not correct it at the source.
Does your PR handle other use cases than the one that would be fixed with #357 ?

thaiphan:
> Does your PR handle other use cases than the one that would be fixed with #357?

No, not off the top of my head.

> Yes the thing is I would rather fix the hook, than adding a logic to "patch the issue" but not correct it at the source.

Well, from a semantic standpoint, you would think that a full re-index would fully re-build your Algolia index (incl. deleting disabled products).

I guess you can look at the Mage ImportExport and see if there's a suitable event somewhere to disable old products upon import.

maxiloc:
Ok makes sense.

As not sure yet of the best option. I've been thinking for a while to implement the full reindex with a temporary index but I am afraid of people doing updates at the same time

thaiphan:
Hey @maxiloc,

I've addressed the issues that you guys brought up. I also added a flag that you can set via the admin panel to turn on the feature. This'll let most people to continue to use the extension without any change in functionality. Those of us who need the feature can just turn it on.

Regards,

Thai

JanPetr:
Hey @thaiphan,
thanks for your work. I took it and little bit updated the code so it's more error proof.
I already merged it here 11fa9ef.

If you have any other ideas or you would like to improve the code, let us know. I will be happy to see another PRs or proposals 💯 

Cheers!

aude:
Just for reference, this feature is also of value to my team. That is because my team has experienced some stores deleting products, but the Algolia module missed that. That might be because of the issue fixed in [1.5.5](https://github.com/algolia/algoliasearch-magento/releases/tag/1.5.5).

So this feature might be necessary for users of <1.5.5 where product delete events were missed.

PR #361, opened 2016-04-21
Set product slaves for all sorting indices when customer groups are enabled

PhilipRowe:
When customer groups are enabled slaves were only being set for price sorts not other sorts that were set such as price, rating etc.


maxiloc:
Oh yes indeed. Nice catch. Thanks for contributing.

PR #362, opened 2016-04-21
Only get attributes for products that are enabled

PhilipRowe:
When adding additional sections like brands some attributes are shown that are only against products that are disabled so will never have any results in the search


maxiloc:
Nice catch again. Thanks

PR #363, opened 2016-04-21
Strip magento block directives

PhilipRowe:
Remove magento specific syntax from indexed pages. When searching for "black..." in the autocomplete dropdown it was showing {{block type="core/template" template="page/html... below the page's name.


maxiloc:
Thanks 😄

maxiloc:
We'll try to do a release by the end of next week with all recent fixes included

PR #366, opened 2016-04-28
CMS content: template directives

t-richards:
# Changed
  - When indexing CMS pages, this module will render template directives by default instead of simply stripping them out.

# Added
  - Adds a configuration option to the Admin Panel which allows the user to change this behavior, if desired.

# Screenshots

Admin Panel:

template-directives


maxiloc:
Thanks for contributing. I will include it in today's release

PR #379, opened 2016-05-13
Save index settings on full products rebuild

JanPetr:
It should fix the bug when you build dev container for the first time and index settings are not set.

_Not sure if we want to call saveSettings everytime when rebuilding products, but I don't know where to put it besides creating a new indexer._


maxiloc:
I think it's good there

PR #380, opened 2016-05-13
Set names to layouts' blocks. Fixes #378

### LINKED ISSUE #378: block name is missing
all the blocks included in layout xml should have name attribute so that it can be customized without modifying the original file. Ex.
`<block type="core/template" template="algoliasearch/beforetopsearch.phtml"/>`

JanPetr:
Not sure if we can use the same block's name (`algolia-beforebodyend`) both in admin and frontend?


maxiloc:
I think it should not be the same

JanPetr:
> I think it should not be the same

Fixed

PR #382, opened 2016-05-13
Code improvements - PSR2, annotations

JanPetr:
@maxiloc I made code_style improvements to follow mostly PSR 2 standarts. I added annotations to most of the class variables so autocomplete can work ;) Please, take a look into that and tell me what you don't like about the code style. We can discuss it and agree on the best solution. Thx ;)

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,17 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->in(__DIR__.DIRECTORY_SEPARATOR.'code');
+
+return Symfony\CS\Config\Config::create()
+    ->setUsingCache(true)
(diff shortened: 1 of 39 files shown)


maxiloc:
Let's setup  StyleCI after this one is merged

JanPetr:
> Let's setup StyleCI after this one is merged

For sure. I can include it even into this PR, hm?

maxiloc:
yes we can do it now

JanPetr:
ok, will do ;) Let me know if you have some issues with the CS I used

maxiloc:
For the CS all for me, let's just make sure we have the same rules than algoliasearch-client-php

JanPetr:
Okey, StyleCI configured and `.php_cs` is configured the same way as in algoliaseach-client-php ;)

PR #386, opened 2016-05-17
Fix big records warning. If there are no long attributes and the size…

JanPetr:
… of the record is high, user is not warned now.


maxiloc:
LGTM, can you just rebase develop so that we can merge

PR #393, opened 2016-05-18
Refactor configuration select fields

### LINKED ISSUE #385: Abstract Admin panel classes
Backport the abstraction did in the magento 2 repo => https://github.com/algolia/algoliasearch-magento-2/blob/master/Model/Source/AbstractTable.php

JanPetr:
Fixes #385 

- All config select fields refactored
- Renamed classes and files
- Abstracted most of logic into parent class

Happy to hear feedback - hit me 👊


code/Block/System/Config/Form/Field/AdditionalSections.php
```diff
@@ -0,0 +1,67 @@
...
+                    'width' => 130,
+                ],
+            ],
+            'buttonLabel' => 'Add Section',
+            'addAfter' => false,
+        ];
+
+        parent::__construct();
+
+        $this->addColumn('label', [
```
maxiloc: Looks better 👍. Why do we have this here instead of in $this->settings['collumns']

JanPetr: I didn't want to mix select fields with text fields. But on the other hand it makes sense to update parent class to handle text fields as well. Will update ;)

code/Block/System/Config/Form/Field/AdditionalSections.php
```diff
@@ -0,0 +1,67 @@
...
+                        
+                        /** @var Algolia_Algoliasearch_Helper_Config $config */
+                        $config = Mage::helper('algoliasearch/config');
+
+                        $attributes = $config->getFacets();
+                        foreach ($attributes as $attribute) {
+                            if ($attribute['attribute'] == 'price') {
+                                continue;
+                            }
+
```
maxiloc: We can merge those two conditions

aude:
Really happy to see your hard work here Jan, thanks!

JanPetr:
Thanks @aude! 
If you have something you would like to improve or fix feel free to open an issue! 🚀

PR #397, opened 2016-05-20
Allow to make facet even from attributes which are not marked as …

### LINKED ISSUE #384: Better handling of facet attributes
- Remove the restriction that a facet needs to be an attribute first
- Merge facets with attributes (https://github.com/algolia/algoliasearch-magento/blob/develop/code/Helper/Config.php#L376-L384)

JanPetr:
… indexible - they will be indexed automatically

Fixes #384

### DIFF
--- a/code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
+++ b/code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
@@ -16,10 +16,10 @@ class Algolia_Algoliasearch_Block_System_Config_Form_Field_ProductAdditionalAttr
 
                         /** @var Algolia_Algoliasearch_Helper_Entity_Producthelper $product_helper */
                         $product_helper = Mage::helper('algoliasearch/entity_producthelper');
-                        
+
                         $searchableAttributes = $product_helper->getAllAttributes();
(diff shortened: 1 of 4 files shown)


code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
```diff
@@ -16,10 +16,10 @@ public function __construct()
 
                         /** @var Algolia_Algoliasearch_Helper_Entity_Producthelper $product_helper */
                         $product_helper = Mage::helper('algoliasearch/entity_producthelper');
-                        
+
                         $searchableAttributes = $product_helper->getAllAttributes();
                         foreach ($searchableAttributes as $key => $label) {
-                            $options[$key] = $key ? $key : $label;
+                            $options[$key] = $key ?: $label;
```
maxiloc: didn't know that one :)

maxiloc:
LGTM

code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
JanPetr: :)

PR #403, opened 2016-05-25
Index SKUs again

aude:
This has happened as a result of a trivial logic error. The commit might not be the most clear way to fix the logic error, feel free to restructure the `if ($value === null || 'sku' == $attribute_name)`, it's complementary `else` and the `if ($type == 'configurable' || ...` otherwise if you can think of a better way.

-------------------------------------------------

6bc7085 wishes to be able to index SKUs for both child products
themselves and for their parent products at the same time. The commit
had the effect of indexing child SKUs for parent products, but child
products themselves did not get indexed any more. This commit enables
indexing of child or standalone product SKUs again.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -816,7 +816,9 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
                             $customData['in_stock'] = 0;
                         }
                     }
-                } else {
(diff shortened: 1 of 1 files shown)


aude:
This might be considered critical/important/master-merge-stuff, as it basically means that 1.5.5 cannot index SKUs for simple products. It's not a problem for <=1.5.4.

aude:
Continued in #405

PR #405, opened 2016-05-26
Fix SKU indexing in simple products

### LINKED ISSUE #404: Algolia 1.5.5, empty SKU for simple product
Hello,
The sku attribute is missing for simple product in the index process.

I make it working by adding this 
`else if('sku' == $attribute_name){
                        $value = $this->getValueOrValueText($product, $attribute_name, $attribute_resource);

                        if ($value)
                        {
                            $customData[$attribute_name] = $value;
                        }
                    }`

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php on line 755

Best,
Florent

JanPetr:
Fixes #404

Hey @aude,
thanks for #403. I took it and did some tests. Unfornutelly it breaks configurable products.

I tried to find the cleanest way how to fix the issue and I move the `$type` condition to previous condition. According my tests it works OK. Could you confirm that? Or do you see any problem in this PR?

Let me know. Once it's OK we will release 1.5.6.

cc @maxiloc

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -712,46 +712,42 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
             {
                 $attribute_resource->setStoreId($product->getStoreId());
 
-                if ($value === null || 'sku' == $attribute_name)
+                if (($value === null || 'sku' == $attribute_name) && ($type == 'configurable' || $type == 'grouped' || $type == 'bundle'))
                 {
-                    /** Get values as array in children */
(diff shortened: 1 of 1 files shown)


aude:
Yes, that's a good idea. I was also considering moving everything into one `if`, was afraid it would be too long, but I think it's alright. Maybe a comment above would be a good idea.

JanPetr:
@aude thank you for your contribution. We decided to release it as part of 1.6.0 probably in the middle of next week.
Now I will close this PR and create a new one targeting `develop` branch.

aude:
Thank you!

Just tested, I can confirm that it works for simple and configurable products here.

PR #408, opened 2016-05-26
Fix category's product count in Algolia index

JanPetr:


code/Helper/Entity/Categoryhelper.php
```diff
@@ -118,8 +118,13 @@ public function getAllAttributes()
...
-        /** @var Mage_Catalog_Model_Resource_Product_Collection $productCollection */
-        $productCollection = $category->getProductCollection();
+        $storeId = $category->getStoreId();
+
+        /** @var Algolia_Algoliasearch_Helper_Entity_Producthelper $productHelper */
+        $productHelper = Mage::helper('algoliasearch/entity_producthelper');
+
+        $productCollection = clone $productHelper->getProductCollectionQuery($storeId);
+        $productCollection = $productCollection->addCategoryFilter($category);
         $productCollection = $productCollection->addMinimalPrice();
```
maxiloc: We can probably remove this one. I don't remember why we have that

JanPetr: Removed. And I also updated call of `getProductCollectionQuery` to not to fetch all data.

PR #409, opened 2016-05-27
Update product index if category products updated

thaiphan:
This change enables the Algolia extension for Magento to now automatically update the product index when a product is added to or removed from a category via the "Manage Categories" menu.

This change also moves some duplicate code into a helper class.

### DIFF
--- a/code/Model/Indexer/Algolia.php
+++ b/code/Model/Indexer/Algolia.php
@@ -1,11 +1,8 @@
 <?php
 
-class Algolia_Algoliasearch_Model_Indexer_Algolia extends Mage_Index_Model_Indexer_Abstract
+class Algolia_Algoliasearch_Model_Indexer_Algolia extends Algolia_Algoliasearch_Model_Indexer_Abstract
 {
     const EVENT_MATCH_RESULT_KEY = 'algoliasearch_match_result';
-
-    /** @var Algolia_Algoliasearch_Model_Resource_Engine */
-    protected $engine;
 
     /** @var Algolia_Algoliasearch_Helper_Config */
     protected $config;
@@ -20,7 +17,6 @@ class Algolia_Algoliasearch_Model_Indexer_Algolia extends Mage_Index_Model_Index
     {
         parent::__construct();
 
-        $this->engine = new Algolia_Algoliasearch_Model_Resource_Engine();
         $this->config = Mage::helper('algoliasearch/config');
         $this->logger = Mage::helper('algoliasearch/logger');
     }
@@ -53,11 +49,6 @@ class Algolia_Algoliasearch_Model_Indexer_Algolia extends Mage_Index_Model_Index
             Mage_Index_Model_Event::TYPE_DELETE,
         ],
     ];
-
-    protected function _getResource()
-    {
-        return Mage::getResourceSingleton('catalogsearch/indexer_fulltext');
-    }
 
     public function getName()
     {
@@ -171,14 +162,6 @@ class Algolia_Algoliasearch_Model_Indexer_Algolia extends Mage_Index_Model_Index
         return $this;
     }
 
-    protected function _isProductComposite($productId)
-    {
-        /** @var $product Mage_Catalog_Model_Product */
-        $product = Mage::getModel('catalog/product')->load($productId);
-
-        return $product->isComposite();
-    }
-
     protected function _processEvent(Mage_Index_Model_Event $event)
     {
         if (!$this->config->getApplicationID() || !$this->config->getAPIKey() || !$this->config->getSearchOnlyAPIKey()) {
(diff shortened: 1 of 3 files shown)


code/Helper/Indexer.php
```diff
@@ -0,0 +1,91 @@
...
+     * This function will update all the requested categories and their child
+     * categories in Algolia. You can provide either a single category ID or an
+     * array of category IDs. A category ID should be either a string or an
+     * integer.
+     *
+     * @param array|string|int $updateCategoryIds
+     */
+    public function reindexSpecificCategories($updateCategoryIds)
+    {
+        $updateCategoryIds = is_array($updateCategoryIds) ? $updateCategoryIds : [$updateCategoryIds];
```
JanPetr: This can be shorten:
```php 
$updateCategoryIds = (array) $updateCategoryIds;
```

code/Helper/Indexer.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return Mage::getResourceSingleton('catalogsearch/indexer_fulltext');
+    }
+
+    /**
+     * @return Algolia_Algoliasearch_Model_Resource_Engine
+     */
+    protected function getEngine()
+    {
+        if (is_null($this->engine)) {
```
JanPetr: I would prefer not to set `NULL` to `$engine` variable and testing `!isset($this->engine)`, but that's only cosmetic update

code/Model/Indexer/Algolia.php
```diff
@@ -13,6 +13,9 @@ class Algolia_Algoliasearch_Model_Indexer_Algolia extends Mage_Index_Model_Index
     /** @var Algolia_Algoliasearch_Helper_Logger */
     protected $logger;
 
+    /** @var Algolia_Algoliasearch_Helper_Indexer */
+    protected $indexer;
```
JanPetr: Woul be better to expand the var name to something like `$indexHelper`. Only `$indexer` might be confusing with `Indexer` classes.

code/Model/Indexer/Algoliacategories.php
```diff
@@ -10,6 +10,9 @@ class Algolia_Algoliasearch_Model_Indexer_Algoliacategories extends Mage_Index_M
     /** @var Algolia_Algoliasearch_Helper_Config */
     protected $config;
 
+    /** @var Algolia_Algoliasearch_Helper_Indexer */
+    protected $indexer;
```
JanPetr: The same with the var name as in `Algolia.php`.

code/Model/Indexer/Algoliacategories.php
```diff
@@ -156,18 +152,15 @@ protected function _processEvent(Mage_Index_Model_Event $event)
...
-                    $updateCategoryIds[] = $category->getId();
-                }
-            }
+            $this->indexer->reindexSpecificCategories($data['catalogsearch_update_category_id']);
+        }
 
-            $this->engine->rebuildCategoryIndex(null, $updateCategoryIds);
+        /**
+         * If we have added any new products to a category then we need t
+         * update that product in Algolia to reflect it.
```
JanPetr: `If we have added any new products to a category then we need to update these products in Algolia indices`
Wouldn't it be better?

code/Model/Indexer/Algolia.php
```diff
@@ -171,14 +175,6 @@ protected function _registerCatalogProductEvent(Mage_Index_Model_Event $event)
         return $this;
     }
 
-    protected function _isProductComposite($productId)
```
JanPetr: I'm not sure we want to remove the method completely. It definitelly makes sense to move the method to one class but I'm worried about such BC break. I am thinking about keeping the method and just call `$this->index->isProductComposite($productId);` from it and trigger `Deprecated` warning? WDYT?

JanPetr:
@thaiphan Thank you for the PR, from my POV makes sense. I just added few comments to some lines. And I will invite @maxiloc to take a look.

JanPetr:
OK, we discussed it and the "helper" approach is not the ideal one. It would be better if you create a new abstract indexer class and move the duplicate code there. It would keep the flow of data and do not cause BC breaks (like with `_isProductComposite` method).

Are you able to refactor it?

thaiphan:
@JanPetr Okay.

thaiphan:
@JanPetr @maxiloc Okay, I updated to use an abstract class. Can you review and merge?

JanPetr:
@thaiphan Look way better, thanks! Can you just squash your commits and fix StyleCI issues?

thaiphan:
@JanPetr I squashed everything but I'm confused about StyleCI. I'm pretty sure I already made the suggested change.

PR #412, opened 2016-05-30
Refactor versioning

JanPetr:
- Refactored retrieval of extension's version - now it's fetched from the unified extension's XML config
- Added correct version to frontend API calls
- Added annotations to `commonjs.phtml` template variables


maxiloc:
👍

PR #413, opened 2016-05-30
Fix already set elements indexing

JanPetr:
When I tried to index attribute `category_ids` it failed because it tried to fetch textual value of attribute. But the attribute was already set. There was probably logical error in condition.
Could you please review as I'm not sure it's not harmful. I tested it a bit and seems to work OK.

### DIFF
--- a/code/Helper/Entity/Producthelper.php
+++ b/code/Helper/Entity/Producthelper.php
@@ -812,17 +812,17 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
                         $customData[$attribute_name] = array_values(array_unique($values));
                     }
 
+                    // Set main product out of stock if all
+                    // sub-products are out of stock.
                     if ($customData['in_stock'] && $all_sub_products_out_of_stock) {
-                        // Set main product out of stock if all
-                        // sub-products is out of stock.
                         $customData['in_stock'] = 0;
                     }
-                } else {
+                } elseif ($value === null) {
                     $value = $this->getValueOrValueText($product, $attribute_name, $attribute_resource);
+                }
 
-                    if ($value) {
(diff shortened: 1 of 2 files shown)


code/Helper/Entity/Producthelper.php
```diff
@@ -817,7 +817,7 @@ public function getObject(Mage_Catalog_Model_Product $product)
                         // sub-products is out of stock.
                         $customData['in_stock'] = 0;
                     }
-                } else {
+                } elseif ($value === null) {
                     $value = $this->getValueOrValueText($product, $attribute_name, $attribute_resource);
```
maxiloc: Seems ok, but we need an else no ? Otherwise the value won't be set in the customData variable

JanPetr: Oh yes, my bad. Will fix

code/Helper/Entity/Producthelper.php
```diff
@@ -812,17 +812,17 @@ public function getObject(Mage_Catalog_Model_Product $product)
...
+                }
 
-                    if ($value) {
-                        $customData[$attribute_name] = $value;
-                    }
+                if ($value && !isset($customData[$attribute_name])) {
```
JanPetr: @maxiloc I didn't add is as else because it would mean some duplicated code once we would have more `elseif`.
But I'm not sure about that. Mainly because `isset`. Is it intentional that the attribute already exists in `$customData` and is in `$additionalAttributes` to be overridden again?

maxiloc: no it's better like this indeed

JanPetr: We are we good to merge? 💥

maxiloc: yep

PR #416, opened 2016-05-30
Do not display categories when instant search is not enabled - it gen…

### LINKED ISSUE #415: Cateories link in the autocomplete
image

We should not display clickable categories if instantsearch is not enabled otherwise the link is wrong

JanPetr:
…erates wrong links

Fixes #415


maxiloc:
👍

PR #433, opened 2016-06-06
160 documentation

JanPetr:


JanPetr:
@maxiloc Could you please review Docs? It should be the same as last time, I only removed Developers page and moved its content to Documentation. And added `Logging` section.

@LukyVj Could you please review my CSS updates? Nothing too serious, just to be sure it's 💯 Thanks!

maxiloc:
👍

PR #434, opened 2016-06-08
[~TASK] Remove absolute skin path on template files

FredericMartinez:
I changed skin static files path to use Magento method `$this->getSkinUrl()`.
It's the proper way to call static files.


JanPetr:
Hi @FredericMartinez,
thanks for PR. Makes sense.

Can you fix it even in [autocomplete.phtml:248](https://github.com/algolia/algoliasearch-magento/blob/develop/design/frontend/template/autocomplete.phtml#L248)?

Thanks!

Issue #424, opened 2016-06-02
Proposal: only IDs in queue


JanPetr:
What if we store in queue only IDs of updated/added/deleted items and then when processing the queue the extension decide what operation it should make.

It will be simpler in case of merging tasks. Only - is ID already in a queue? 
- Yes, ok do not insert
- No, ok, insert into queue table

And it would decrease operations count on indexing part.
On processing the extension will retrieve IDs from DB on depending on current state of item (product, category, page, ...) will decide what operation it should push to Algolia API. So instead of pushing `add - delete - add` it will push only `add` because it's the current state of the item.

maxiloc:
Yes that would work, the only thing is we still have actions that are not ID related (full reindex, setsettings, move).

JanPetr:
Yes, that shouldn't be a problem to keep this "mass" operations in the queue as well and process them. I will try to make some PoC.

PR #438, opened 2016-06-10
Simpler products removing

JanPetr:
Implements #424 

- works everytime
- no specific configuration needed
- keeps indices in consistent state
- simplifies code
- makes queue more effective, because there is no need to add `removeProducts` to queue table. So only merged `rebuildProducts` tasks are processed.

### DIFF
--- a/code/Helper/Data.php
+++ b/code/Helper/Data.php
@@ -69,7 +69,7 @@ class Algolia_Algoliasearch_Helper_Data extends Mage_Core_Helper_Abstract
         $this->algolia_helper->deleteIndex($this->category_helper->getIndexName($storeId));
     }
 
-    public function saveConfigurationToAlgolia($storeId)
+    public function saveConfigurationToAlgolia($storeId, $saveToTmpIndicesToo = false)
     {
         $this->algolia_helper->resetCredentialsFromConfig();
 
@@ -99,7 +99,7 @@ class Algolia_Algoliasearch_Helper_Data extends Mage_Core_Helper_Abstract
                 $this->additionalsections_helper->getIndexSettings($storeId));
         }
 
-        $this->product_helper->setSettings($storeId);
+        $this->product_helper->setSettings($storeId, $saveToTmpIndicesToo);
     }
 
     public function getSearchResult($query, $storeId)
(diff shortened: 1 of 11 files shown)


code/Helper/Entity/Producthelper.php
```diff
@@ -139,8 +136,13 @@ public function getProductCollectionQuery($storeId, $productIds = null, $only_vi
                 $catalogInventory_stock->addInStockFilterToCollection($products);
             }
 
-            $products = $products->addAttributeToSelect('special_from_date')->addAttributeToSelect('special_to_date')
-                             ->addAttributeToFilter('status', Mage_Catalog_Model_Product_Status::STATUS_ENABLED);
+            // $products = $products->addFinalPrice();
```
JanPetr: @maxiloc I checked and `addFinalPrice()` method only adds some new `price` attributes to `Product` objects. But do not overrides `price` attribute. And we use only `price` attribute for inidicing. So according my research we should not need to call `addFinalPrice()` at all.
And I couldn't find a usecase where we would need it.

Did I miss something?

maxiloc: Never sure about this one :).

We need to test every combinaison before removing it
- price
- price + special price
- price + group price
- price + catalog rule
- price + group price + special price
- price + catalog rule + special price
- price + catolog rule + group price

JanPetr: OK, my bad. We need `addFinalPrice()` 😢

JanPetr:
When queue is enabled, products are re-indexed using temporary index

code/Helper/Data.php
```diff
@@ -312,20 +304,24 @@ public function rebuildStoreProductIndex($storeId, $productIds)
...
 
         try {
-            $collection = $this->product_helper->getProductCollectionQuery($storeId, $productIds);
-
+            $collection = $this->product_helper->getProductCollectionQuery($storeId, $productIds, false);
             $size = $collection->getSize();
 
+            if (!empty($productIds)) {
+                $size = max(count($productIds), $size);
+            }
```
maxiloc: why do we have this ?

JanPetr: Because there might be IDs of deleted products which will never be in `$collection`

code/Helper/Data.php
```diff
@@ -501,13 +519,22 @@ public function rebuildStoreProductIndexPage($storeId, $collectionDefault, $page
         $this->logger->stop('LOADING '.$this->logger->getStoreName($storeId).' collection page '.$page.', pageSize '.$pageSize);
 
         $index_name = $this->product_helper->getIndexName($storeId);
+        if ($useTmpIndex === true) {
+            $index_name .= '_tmp';
+        }
```
maxiloc: I would put that directly in getIndexName

code/Helper/Data.php
maxiloc: 👍

maxiloc:
Looks good

code/Helper/Data.php
JanPetr: Refactored

Issue #426, opened 2016-06-02
Logging - show deletes


sam327:
When I enable the log file in magento, I see lots and lots of create record items in algolia.log, but I don't see any references to items being _deleted_ from the algolia index. 

Like I see bunches of: ```>>>>> BEGIN CREATE RECORD``` & ```>>>>> BEGIN SEND TO ALGOLIA```

But I also expect to see a bunch of ```>>>>> BEGIN DELETE RECORD``` and I don't see any. Are deletes just not getting logged? Or am I looking at the wrong thing?

I'm trying to match up a set of recent transactions from algolia's website with those in the magento file.

I'm still on 1.5.3.

JanPetr:
Hello @sam327,
thanks for your question. Right now we do not log delete operations so you cannot see them in log file.

I let this issue open as it might be something we want to add. If you want to add it yourself feel free to send PR :)

sam327:
Ok - more important than anything else was making sure I was reading things right. What does it track? Having a list would be a nice addition to the instructions. 

And then figuring out what it _should_ track is another thing entirely.

It looks like the recent transactions on algolia.com only get saved for a day (or less), and you can't download them in bulk. So the lack of quality logs is frustrating.

JanPetr:
Hey @sam327, I understand.
We will improve logging in future versions.

Here is a short list of what is logged now:
- Full product reindex
- Rebuilding one page of products
- Loading products' collection
- Creating of products' records
- Creating of single product's record
- Start and stop of sending objects to Algolia
- Start and stop of emulation
- Exceptions from images' loading
- Miscellaneous errors and exceptions

JanPetr:
Will be part of the next release

PR #442, opened 2016-06-15
Log updated and deleted IDs

JanPetr:
I know it will produced big data, but it might be useful for debugging. WDYT @maxiloc?

Implements #426


maxiloc:
Let's go for it

Issue #431, opened 2016-06-06
To add algoliasearch.csv into /locales/en_US


tomekjordan:
Would be great to add en_US locales to fulfill with translation strings. Some strings are very hard to translate. So would be great if you include own as reference.

JanPetr:
Implemented, it will be part of the next release.

PR #446, opened 2016-06-20
Add locale CSV file

JanPetr:
Implements #431


JanPetr:
I'm not sure if to put the CSV into `en_GB` or `en_US`. Magento in our Docker container has default language set to `en_GB`, but I don't know globally. Do you have any insights @maxiloc?

maxiloc:
Not really, I guess both could do.

JanPetr:
OK, let's keep GB.

PR #450, opened 2016-06-21
Fix indexing queue data

JanPetr:


code/sql/algoliasearch_setup/mysql4-upgrade-1.5.0-1.6.0RC.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
```
maxiloc: We need to be careful with the filename, because I am not sure what happens when you release 1.6.0

maxiloc: otherwise 👍

maxiloc: In fact I would rather empty the queue

JanPetr: > We need to be careful with the filename, because I am not sure what happens when you release 1.6.0

With releasing 1.6.0 I want to rename the file. There might be an issue that 2 same files might be there - RC and non-RC. Question is how big issue is that or what other options we have?

> In fact I would rather empty the queue

I was thinking about that but if you empty the queue completly you would loose operations. On the other hand, you must force complete re-index because of the BC breaks. Yes. I will update it to deleting all rows from queue.

code/sql/algoliasearch_setup/mysql4-upgrade-1.5.0-1.6.0RC.php
```diff
@@ -0,0 +1,11 @@
...
+/** @var Mage_Core_Model_Resource_Setup $installer */
+$installer = $this;
+
+$installer->startSetup();
+
+// Need to transform "removeProducts" to "rebuildProductIndex" as re-indexing was refactored and "removeProducts" do not exists anymore
+$installer->run("DELETE FROM `{$installer->getTable('algoliasearch/queue')}`");
```
JanPetr: I changed the query? Do you have any idea what to do with the filename? ^ @maxiloc

PR #451, opened 2016-06-21
Fix showing instant search bars on products' details

JanPetr:
@maxiloc It was caused because instant search was no started. I strart it everytime now and works. I tested all the combinationas turned on/off autocomplete / instant search and it works OK. Did I miss some use case or can we leave it like that?

### DIFF
--- a/design/frontend/template/instantsearch.phtml
+++ b/design/frontend/template/instantsearch.phtml
@@ -566,24 +566,12 @@ if ($config->isInstantEnabled() && $config->replaceCategories() && Mage::app()->
                 })
             );
 
-            /** Initialise searching **/
-            if (algoliaConfig.isSearchPage) {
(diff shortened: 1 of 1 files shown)


design/frontend/template/instantsearch.phtml
```diff
@@ -567,23 +567,14 @@ if ($config->isInstantEnabled() && $config->replaceCategories() && Mage::app()->
...
-
-                    $(instant_selector).one('focus', (function () {
-                        if (!isStarted) {
-                            startInstantSearch();
-                        }
-                        isStarted = true;
-                    }));
-                })();
-            }
+            startInstantSearch();
```
maxiloc: let put this one after the declaration for ie

maxiloc:
👍

PR #453, opened 2016-06-22
Fix content restoration on cross click

### LINKED ISSUE #452: Clicking the cross in the search field does not return the wrapped content
This comes off the back of #302 , but I thought it was probably worth creating a new issue.

When instant search is on and autocomplete is off, typing into the search field replaces the page content with search results (as expected), and clearing the values from the search field returns the page content (as expected). Clicking the cross button to clear the search does not return the page content, however, and the search results remain until you type something in the search field and then remove it manually.

JanPetr:
Fixes #452


maxiloc:
👍

PR #456, opened 2016-06-23
Fix indexed values of selectable attributes

JanPetr:


code/Helper/Entity/Producthelper.php
```diff
@@ -859,7 +860,7 @@ public function getObject(Mage_Catalog_Model_Product $product)
                     if ($customData['in_stock'] && $all_sub_products_out_of_stock) {
                         $customData['in_stock'] = 0;
                     }
-                } elseif ($value === null) {
+                } elseif (!is_array($value)) {
```
maxiloc: We might want an OR there, not sure it's an array everytime ?

maxiloc: The format is different depending on the attribute type

JanPetr: Well, yes, but all other format will pass except `array`. So even `NULL` will go through. Only `array` won't.

maxiloc: OK it make sense then

PR #459, opened 2016-06-24
Now when you re-index suggestions it deletes it's settings. This comm…

JanPetr:
…it fixes that.


maxiloc:
👍

JanPetr:
Not really :)) I found a better solution.

maxiloc:
It's ok to set settings at each reindex

JanPetr:
Yes, but it's not OK to set settings to full index. It's always better to set it to empty `tmp` index :)

JanPetr:
Now you can review :)

maxiloc:
ah yes 😄

PR #461, opened 2016-06-27
Create release package automatically

JanPetr:


dev/bin/makeRelease.php
```diff
@@ -0,0 +1,27 @@
...
+/** @var Algolia_Algoliasearch_Helper_Config $config */
+$config = Mage::helper('algoliasearch/config');
+
+$data = $connectHelper->loadLocalPackage('algoliasearch');
+$data['version'] = $config->getExtensionVersion();
+
+$model->setData($data);
+
+if ($model->createPackage()) {
+    echo 'Release package was successfully created.';
```
maxiloc: Not sure what can happen but could be nice to have explicit message in case of failure

maxiloc:
Really nice 👍

dev/bin/makeRelease.php
```diff
@@ -0,0 +1,27 @@
...
+    exit(0);
+}
+
+echo 'Release package could not be created.';
```
JanPetr: Here is the explicit message. But it will alway be shown inside the container. Don't know how can I display in local terminal as this script runs inside the container :(

maxiloc: This should do for printing, but I was more speaking about a message "wrong permission on ... folder" or "memory issue while creating the package", not sure what can happen and what we can catch from magento

JanPetr: Oh, now I got it. We can catch any exception on that level. Magento catches everything in higher level.
But now sure if it makes sense to catch it. If it fails, PHP itself will output the error message, no?

maxiloc: Not sure, we had permission issue when we tried it, but we did not get any output. Probably we can iterate on that if we encounter any issue, probably it can be fixed at the container level

JanPetr: Yes, I tested it on local machine. We can test it tommorow on demos server ...

PR #462, opened 2016-06-27
Apply "Show categories that are not included in the navigation menu" …

### LINKED ISSUE #460: The "Show categories that are not included in the navigation menu" is only applied to autocomplete
It should be applied to instant search as well

JanPetr:
…setting to products' categories so it's applied even on instant search pages

Fixes #460


maxiloc:
👍

PR #463, opened 2016-06-28
Prices being indexed excl.tax instead of incl.tax in some tax setups

Floddy:
In some setups, like when catalog prices are maintained excl. tax and displayed in the store with tax, the price still get indexed without tax, even when $with_tax returns true here (Helper/Entity/Producthelper.php, line 315):

$price = (double) Mage::helper('tax')->getPrice($product, $product->getPrice(), $with_tax, ...

This seems to be due to that the tax helper's getPrice function tries to get the tax percant / tax class from $product, but $product doesn't include that, returns NULL, and no tax is added.

This fixes that by including the tax class id in the product collection.


JanPetr:
Nice catch, thank you for contributing @Floddy 💯

PR #471, opened 2016-06-28
Fixed PHP 5.3 syntax

redox:


JanPetr:
Hey @redox,
in Magento extension we do not support PHP 5.3, because even Magento 1.6.2 do not support it. So there is no reason to keep an old array syntax.
Thank you anyway :)

redox:
> Hey @redox,
> in Magento extension we do not support PHP 5.3, because even Magento 1.6.2 do not support it. So there is no reason to keep an old array syntax.
> Thank you anyway :)

Ah OK, got confused by the README but it seems you've fixed it: 66cc49fa7081f5ee52ade5770cacc40a0e516676

PR #488, opened 2016-07-11
Escape translated JS strings

JanPetr:


maxiloc:
why did you not put true on every translation ?

JanPetr:
Because `json_encode` is proper way how to escape JS strings. `addslashes` needs to be used only when string is injected to another string (HTML, ...) because `json_encode` don't work there as expected.
But when you have pure string put into JS var, `json_encode` is the best way to go.

maxiloc:
ok make sense 👍

PR #493, opened 2016-07-13
Send more information in UA

JanPetr:
Those are the information we ask in 90% of support tickets.
Dashboard and other system will have no problem with the UA format, checked.


maxiloc:
👍

Issue #487, opened 2016-07-11
Optimize the size of inline generated JS code


JanPetr:

JanPetr:
Will be part of the next release.

PR #507, opened 2016-07-18
Optimize frontend

JanPetr:
- ~~Completely removed dependency on PHP from JS templates~~
- Separated JS code to separated JS files
- Only `algoliaConfig` variable and JS templates are now inlined to HTML
- JS files are loaded properly via Magento's "addJs" method so it supports merging of scripts
- JS templates are still in PHTML files to be easily overridable
- JS files structured into proper directories
- JS code restructuralized a bit

Implements #487


design/frontend/template/autocomplete/menu.phtml
```diff
@@ -7,9 +7,16 @@
         <div class="col3">
             <div class="other-sections">
                 <div class="aa-dataset-suggestions"></div>
-                <?php for ($i = 0; $i < 10; $i++): ?>
-                    <div class="aa-dataset-<?php echo $i; ?>"></div>
-                <?php endfor; ?>
+                <div class="aa-dataset-0"></div>
```
maxiloc: We can keep the loop, since it's still in a .phtml

design/frontend/template/autocomplete/product.phtml
```diff
@@ -1,32 +1,16 @@
...
+        <div class="thumb"><img src="{{{ baseUrl }}}/{{ thumbnail_url }}" /></div>
         {{/thumbnail_url}}
 
         <div class="info">
             {{{_highlightResult.name.value}}}
 
             <div class="algoliasearch-autocomplete-category">
                 {{#categories_without_path}}
-                <?php echo $this->__('in'); ?> {{{categories_without_path}}}
+                {{ inTranslation }} {{{categories_without_path}}}
```
maxiloc: For translation I would keep the php method, it's easier to customize

JanPetr: Well I wanted to remove all PHP dependency as first intention was put everything to JS files. There is bunch of similar occurences of that. So now it depends how we decide to proceed.

js/commonJs.js
```diff
@@ -0,0 +1,462 @@
+document.addEventListener("DOMContentLoaded", function (e) {
```
maxiloc: I would also go for an internal folder for js

maxiloc:
Looking good

js/commonJs.js
JanPetr: Yes, good idea. But once we are on this file - I was thinking about moving "createAutocompleteSource" and "addFacetWidget" method to instantsearch, respectively autocomplete.js file. Because logically it belogs there. WDYT?

design/frontend/template/autocomplete/product.phtml
maxiloc: For me it makes sense to keep them. Your call on this one

design/frontend/template/autocomplete/menu.phtml
JanPetr: That's the question where to stop. If we want to keep PHP or not. The same stuff as with translations in the templates ...

js/commonJs.js
maxiloc: It's tricky, the goal to separate them was to keep the file you update regurlarly when doing implem of algolia really minimal.
Most of the code is not really understandable by everyone, I would keep it separate

JanPetr: You mean those two methods, right?
It happend to me couple of times that I linked the code of `addFacetWidget` method to customers because they wanted to update behaviour of some widgets. All widgets are loaded in instatnsearch.js file except for facets. That's why I would move the method - it's updated by customers and "internals" should not be customized.

`getAutocompleteSource` is not that obvious, true.

PR #509, opened 2016-07-20
FAQ entry about why some products do not come up when searching

JanPetr:


maxiloc:
Nice I would add a small part with a link to the collection fetching

JanPetr:
Good idea, I added a small part to docs as well ;)

redox:
Awesome @JanPetr 👍 

(cc @clement-leprovost)

PR #518, opened 2016-07-27
Fix the issue with CDN images

### LINKED ISSUE #515: Hit templates use Mage::getBaseUrl instead of getting the media url.
You can't use a CDN for images with this extension because the url isn't the Magento media url configuration value.

JanPetr:
Fixes #515

### DIFF
--- a/code/Helper/Entity/Categoryhelper.php
+++ b/code/Helper/Entity/Categoryhelper.php
@@ -69,6 +69,8 @@ class Algolia_Algoliasearch_Helper_Entity_Categoryhelper extends Algolia_Algolia
         }
 
         $additionalAttr[] = 'include_in_menu';
+        $additionalAttr[] = 'image';
+        $additionalAttr[] = 'thumbnail';
 
         $categories
             ->addPathFilter($storeRootCategoryPath)
@@ -143,11 +145,6 @@ class Algolia_Algoliasearch_Helper_Entity_Categoryhelper extends Algolia_Algolia
             $path .= $this->getCategoryName($categoryId, $storeId);
         }
 
-        $image_url = null;
-        try {
-            $image_url = $category->getImageUrl();
-        } catch (Exception $e) { /* no image, no default: not fatal */
-        }
         $data = [
             'objectID'        => $category->getId(),
             'name'            => $category->getName(),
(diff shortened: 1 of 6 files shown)


code/Helper/Entity/Categoryhelper.php
```diff
@@ -118,6 +118,8 @@ public function getAllAttributes()
 
     public function getObject(Mage_Catalog_Model_Category $category)
     {
+        $category->load($category->getId());
```
Inwerpsel: I fixed the image field not being available when doing $category->getImageUrl() by adding the attribute 'image' in getCategoryCollectionQuery(). This way you can get the image without having to load each category of the collection again.

code/Helper/Image.php
```diff
@@ -31,16 +31,11 @@ public function toString()
...
-    public function getUrl(Mage_Catalog_Model_Product_Image $model)
+    public function removeProtocol($url)
     {
-        $baseDir = Mage::getBaseDir();
-        $path = str_replace($baseDir.DS, '', $model->getNewFile());
-
-        $url = str_replace(DS, '/', $path);
-
-        return $url;
+        return str_replace(['https://', 'http://'], '//', $url);
```
Inwerpsel: I solved it (at work, that why I couldn't share it yet) by removing 'media' from $url and then passing the configuration media url along with the hit template.

Like this:  $path = str_replace($baseDir . DS . 'media' . DS, '', $model->getNewFile());

That way you don't have to store the same media url on each record for products and categories.

JanPetr: Yes, we had the same thing in version 1.5.5. But it ended up in issues with placeholder images.
Indexing whole URL is the most straight forward and the most understandable approach as the logic is only at one place and not spread between indexing and front-end part.

code/Helper/Entity/Categoryhelper.php
JanPetr: Good idea, will fix! Thanks!

JanPetr: Fixed

code/Helper/Image.php
Inwerpsel: That seems more robust indeed.

design/frontend/template/autocomplete/category.phtml
```diff
@@ -1,37 +1,26 @@
...
-
 <!-- Category hit template -->
 <script type="text/template" id="autocomplete_categories_template">
     <a class="algoliasearch-autocomplete-hit" href="{{url}}">
         {{#image_url}}
-        <div class="thumb">
-            <img src="<?php echo $image_base_url; ?>{{image_url}}" />
-        </div>
+            <div class="thumb">
+                <img src="{{image_url}}" />
```
Inwerpsel: It would be nice to be able to use the category thumbnail image for this, since category images cannot be resized and are therefor many times larger than the small size needed for the category suggestions.

It is not possible however to add these in the configuration so I had to add these with code (add attribute 'thumbnail' to loaded category collection).
It would be nice if the thumbnail field was fetched by default or if there was an option to add it to the indexed fields.

code/Helper/Entity/Categoryhelper.php
```diff
@@ -160,8 +157,16 @@ public function getObject(Mage_Catalog_Model_Category $category)
             'product_count'   => $category->getProductCount(),
         ];
 
-        if (!empty($image_url)) {
-            $data['image_url'] = $image_url;
+        try {
+            $imageUrl = $this->getThumbnailUrl($category) ?: $category->getImageUrl();
```
JanPetr: cc @Inwerpsel
Makes sense to you? Indexing thumbnail when it's available. Otherwise index main image.

Inwerpsel: Yes, what I had was index both but this might be better since this module only needs one small image on the frontend.

PR #524, opened 2016-08-01
Restore PHP 5.3 version

### LINKED ISSUE #523: Restore PHP 5.3 compatibility
Magento's versions 1.8.X and lower still supports PHP version 5.2. 
As support PHP 5.2 does not make much sense anymore, PHP 5.3 still needs to be supported by the extension.

JanPetr:
Fixes #523, https://magento.ugc.bazaarvoice.com/4643-en_us/27618/review/150624401/redirect.htm

cc @redox

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,14 +1,14 @@
 <?php
 
 $finder = Symfony\CS\Finder\DefaultFinder::create()
-    ->in(__DIR__.DIRECTORY_SEPARATOR.'code');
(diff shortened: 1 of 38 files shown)


maxiloc:
👍, I don't know if we can respond to the message on connect. Because it would be good to say that we fixed it and that using 5.3 is a security issue on their side

redox:
Great 👍  I hope you have tools to make such changes automatic :)

JanPetr:
> I hope you have tools to make such changes automatic :)

I do, I'd die doing it manually :-D

PR #534, opened 2016-08-08
Encode attribute values

Zifius:
We had this problem when manufacturer attribute contained "+" or "&" characters, in this case in suggestion list the url is not escaped leading to special symbols being dropped and no results are returned, specially with "&" only symbols before are used for search. Yet after this change the selected filters have the ampersand double escaped, but at least there are results for the search


JanPetr:
Makes sense, thanks for contributing! 💯

Zifius:
My pleasure :)

PR #548, opened 2016-08-17
Removed the No or N/A values from being indexed

royduin:
If a product attribute is empty, Magento is returning N/A or No. By excluding these values we get cleaner facets instead of a lot of values in a facet with just one option: N/A.

### DIFF
--- a/code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
+++ b/code/Block/System/Config/Form/Field/ProductAdditionalAttributes.php
@@ -51,6 +51,14 @@ class Algolia_Algoliasearch_Block_System_Config_Form_Field_ProductAdditionalAttr
                     ],
                     'rowMethod' => 'getOrder',
                 ],
+                'index_no_value' => [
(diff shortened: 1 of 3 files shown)


JanPetr:
Hello @royduin,
thanks for contribution. Unfortunatelly this is not something we can add to the extension like that as it breaks backward compatibility and now all attributes with value "No" will be removed from index. And that will cause troubles with people who want to index their "No" attributes.
Can you make it configurable with default value set on off?

royduin:
Something like this @JanPetr?

JanPetr:
Exactly @royduin, thanks!
Just two things - can you make the header shorter and explain what the settings mean below in comments of the field. Something like that:
screenshot 2016-08-19 10 42 45

And can you develop it on top of `develop` branch instead of the `master` one? And target the PR on `develop` as well?

Thanks!

royduin:
Done, see: https://github.com/algolia/algoliasearch-magento/pull/554

Issue #322, opened 2016-02-19
hits widget emits divs


rbrown:
Hi, the hits widget emits a couple of divs around each hit which is breaking the layout that I'm trying to do for my results page.

```HTML
<div class="ais-hits" data-reactid=".1">
<div class="ais-hits--item" data-reactid=".1.$3953752">
```

I can't see how to disable the js from doing this? I see there's a new allItems template option in  https://github.com/algolia/instantsearch.js/commit/1f3f889d4cc246c317d9b0e7b436f5d427dc6c33 but that's not in the magento extension yet. Would using the newer version be the best solution for me?

vvo:
Yes @rbrown the new allItems option might be good for you. It will be available in magento soon, stay tuned here by subscribing to the issue (you already are subscribed in fact)

rbrown:
Great, thanks

PR #553, opened 2016-08-19
Hits widget: Use allItems template not item

rbrown:
Following on from issue #322 this patch updates the hit widget to use the
allItems template. It updates hit.phtml with the {{hits}} tags so that it
continues to work as before.

The diff on hit.phtml is a lot more readable if you diff with -b


maxiloc:
@rbrown Thanks for contributing, It looks good to me. I'll let @JanPetr review it before merging

JanPetr:
Looks good! Thanks for contributing @rbrown

PR #576, opened 2016-09-12
Hide IS page info panel on no results

### LINKED ISSUE #538: When no results are displayed on IS page "SORT BY" is still visible and should not be
screenshot 2016-08-09 18 02 37

JanPetr:
Fixes #538

### DIFF
--- a/js/algoliasearch/instantsearch.js
+++ b/js/algoliasearch/instantsearch.js
@@ -191,6 +191,8 @@ document.addEventListener("DOMContentLoaded", function (event) {
 				}
 			},
 			render: function (data) {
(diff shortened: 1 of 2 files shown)


maxiloc:
What's the issue showing it when there is no result ?

JanPetr:
It does not look nice, it's empty on it's left side and on the right side there is a "sort" select box but you have nothing to sort ...

maxiloc:
Indeed makes sense, I usually prefer to hide/show with a class, but that's a personal thing 😄 .
LGTM

JanPetr:
Class is definitelly better and easier to override. Will change, thanks for the tip @maxiloc

PR #579, opened 2016-09-13
Less aggressive "new version" message

### LINKED ISSUE #544: Make "new version" note less aggressive as it feels more like an error than just a kind reminder

JanPetr:
+ Added small line with rate request

Fixes #544


maxiloc:
👍

PR #580, opened 2016-09-13
Don't include some bundled products' attributes within its' master product

### LINKED ISSUE #526: Bundle products issue with special price and new from/to dates
I have a bundle product with no special price and no new from / to dates configured. In this bundle product are two other products A and B. Product A has a special price configured and product B the new from / to dates. When I take a look at the index for the bundle product it has the special price value from product A and the new from / to dates from product B.

I expected that the bundle has none of the values of product A and B. Can you confirm this issue? 

I am using version 1.5.5.

JanPetr:
Fixes #526 

Really not sure about the approach. Any feedback appreciated.


maxiloc:
Ok with the fix. Just need to make sure it does not break when child product are set to be visible individually

JanPetr:
It's OK, it only skips the attributes when they are fetched from sub-products. When it's master product (and product visible individually is master product) it works just fine ;)

Issue #532, opened 2016-08-08
Add attribute to product with category levels without path


JanPetr:
Right now the `categories` attribute looks like this:
```json
categories: {
    "level0": ["Men"], 
    "level1": ["Men /// Suits", "Men /// Accessories"]
}
```

The idea is to add onther attribute, lets call it `main_categories`, which would contain categories like:
```json
main_categories: {
    "level0": ["Men"], 
    "level1": ["Suits", "Accessories"]
}
```
It provides more flexibility with relevancy and allow to provide more relevant results.

maxiloc:
same here, it will probably break the hierarchical widget

JanPetr:
It'd be added as completely new attribute (`main_categories`), so hierarchical (oh, that's hard to spell 😄) widget should be fine.

JanPetr:
Will be part of the next release

PR #581, opened 2016-09-13
Index hierarchical categories without path in new main_categories attribute

JanPetr:
Adds more relevancy options in category listings with Algolia

Implements #532


maxiloc:
I need to see this one in action.

JanPetr:
Go ahead and test it. It works the same way as described in #532

Issue #533, opened 2016-08-08
Add possibility to index only categories where product is


JanPetr:
Right now when there is category path: "Men /// Blazers" and product is added only to Blazers category the extension indexes:
```json
categories: {
    "level0": ["Men"], 
    "level1": ["Men /// Blazers"]
}
```
It ends up with issue that products which are supposed to be only in Blazers category are fetched even in "Men" category.

It'd be great to add a configuration option to index only categories into which the product actually belong and not the whole category path.

maxiloc:
We need to be careful with this, it might break the hierarchical widget, but yes makes sense

JanPetr:
Yes, that needs to be tested well. If the product is only in "Blazers" category I think that this
should work:
```json
categories: {
    "level0": [], 
    "level1": ["Men /// Blazers"]
}
```

JanPetr:
will be part of the next release

PR #584, opened 2016-09-13
Add option to choose where index or not the whole category tree of the product

JanPetr:
Implements #533 

Breaks hierachical widget of IS page. It's mentioned in comment of configuration input.

### DIFF
--- a/app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
+++ b/app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
@@ -118,6 +118,17 @@ if ($config->isInstantEnabled()) {
 
         $isSearchPage = true;
     }
+
+    /** @var Algolia_Algoliasearch_Helper_Logger $logger */
+    $logger = Mage::helper('algoliasearch/logger');
+    if ($logger->isEnable() && $config->indexWholeCategoryTree() === false) {
+        foreach ($config->getFacets() as $facet) {
+            if ($facet['attribute'] === 'categories') {
+                $logger->log('!! Warning !! You have enabled categories\' hierarchical widget but you do not index whole categories tree. The widget does not display correct data.');
+                break;
+            }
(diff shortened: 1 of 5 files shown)


maxiloc:
Is there a point to add this if it breaks the hierarchical faceting ? because in the end the issue was for the hierarchical widget. Using a refinementList with categories_without_path already works no ?

JanPetr:
Yes, I got customer who didn't use hierachical widget and wanted filtering by categories. Because `categories_without_path` does not contain leveling and are without a path. So when you are in `Blazers /// Men` you would not be able to filter it as `categories_without_path` contains only "Blazers".

maxiloc:
So for only for "Replace category page" ?

JanPetr:
Yes, the issue is described in #533

maxiloc:
Ok for me, we should add some kind of log message in case you put false + hierarchical category widget

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -118,6 +118,17 @@ if ($config->isInstantEnabled()) {
...
+    /** @var Algolia_Algoliasearch_Helper_Logger $logger */
+    $logger = Mage::helper('algoliasearch/logger');
+    if ($logger->isEnable() && $config->indexWholeCategoryTree() === false) {
+        foreach ($config->getFacets() as $facet) {
+            if ($facet['attribute'] === 'categories') {
+                $logger->log('!! Warning !! You have enabled categories\' hierarchical widget but you do not index whole categories tree. The widget does not display correct data.');
+                break;
+            }
+        }
+    }
```
JanPetr: @maxiloc? OK?

maxiloc: yes perfect

Issue #577, opened 2016-09-13
Add option to disable overriding synonyms from Magento


JanPetr:
There are some users who uploads CSVs or are used to set synonyms directly in Algolia dashboard. To keep their comfort it'd be nice to implement simple checkbox in Synonyms tab which would dis/enable the setting (and overriding) synonyms from Magento

ianjs:
+1 

We upload synonyms direct from Algolia via CSV and they mysteriously vanished after we upgraded the module.

One workaround for us would have been to start uploading them from Magento instead, but it looks like CSV file import is not available. It would be nice if this was also supported in the module so our staff can stay in the Magento backend to maintain Algolia.

JanPetr:
Will be part of the next release

PR #585, opened 2016-09-14
Add option to disable synonyms management in Magento

JanPetr:
Implements #577

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/etc/system.xml
+++ b/app/code/community/Algolia/Algoliasearch/etc/system.xml
@@ -648,6 +648,20 @@
                         ]]>
                     </comment>
                     <fields>
+                        <enable_synonyms translate="label comment">
+                            <label>Enable Synonyms management</label>
+                            <frontend_type>select</frontend_type>
+                            <source_model>adminhtml/system_config_source_yesno</source_model>
+                            <sort_order>0</sort_order>
+                            <show_in_default>1</show_in_default>
+                            <show_in_website>1</show_in_website>
+                            <show_in_store>1</show_in_store>
+                            <comment>
+                                <![CDATA[
+                                If set to No, synonyms will be managed only in Algolia's dashboard and won't get overriden by Magento
+                                ]]>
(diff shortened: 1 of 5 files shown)


app/code/community/Algolia/Algoliasearch/etc/system.xml
```diff
@@ -648,6 +648,20 @@
...
+                            <show_in_website>1</show_in_website>
+                            <show_in_store>1</show_in_store>
+                            <comment>
+                                <![CDATA[
+                                If set to No, synonyms will be managed only in Algolia's dashboard and won't get overriden by Magento
+                                ]]>
+                            </comment>
+                        </enable_synonyms>
                         <synonyms translate="label comment">
                             <label>Synonyms</label>
```
maxiloc: I would add a dependency for other synonyms settings, to not display when synonyms is False

js/algoliasearch/internals/adminhtml/admin_scripts.js
```diff
@@ -54,4 +54,24 @@ algoliaAdminBundle.$(function($) {
 			return parseInt(n, 10);
 		});
 	}
+
```
maxiloc: All that should be handled by the dependency

app/code/community/Algolia/Algoliasearch/etc/system.xml
JanPetr: I did that. But it does not work on `synonyms` and `one-way synonyms`, it works only on `synonyms file`. I don't know why, probably Magento's bug. I wrote own script for that: https://github.com/algolia/algoliasearch-magento/pull/585/files#diff-a7e898950149f6c176346608d6d5cba0R58

js/algoliasearch/internals/adminhtml/admin_scripts.js
JanPetr: See https://github.com/algolia/algoliasearch-magento/pull/585#discussion_r78739478 :)

PR #596, opened 2016-09-22
Change algolia queue data column type to TEXT

VincentMarmiesse:
Hello,

The `data` column of the queue table (`algoliasearch_queue`) is a `VARCHAR(5000)` so if you reindex more than 1000 products for example, it will cut the json and make it invalid.

This PR updates the type to `TEXT`.


JanPetr:
Thanks @VincentMarmiesse

PR #599, opened 2016-09-26
Do not process duplicate products

JanPetr:


maxiloc:
Why does this happens ?

JanPetr:
https://github.com/algolia/algoliasearch-magento/blob/master/app/code/community/Algolia/Algoliasearch/Model/Indexer/Abstract.php#L55-L63
2 different configurable products can be composed by same simple products

maxiloc:
Maybe if would be better to put the unique there no ?

JanPetr:
I put it lower as the method can be called from other places then just Abstract indexer and there is the last spot before batching.

maxiloc:
Ok works for me

PR #603, opened 2016-09-29
Add option to write custom attribute name used for custom ranking

JanPetr:


js/algoliasearch/internals/adminhtml/admin_scripts.js
```diff
@@ -75,4 +75,33 @@ algoliaAdminBundle.$(function($) {
 			$synonymsRows.hide();
 		}
 	}
+
```
maxiloc: Why is this doing ?

JanPetr: It shows the text input field for custom attribute only when you have selected `[use custom attribute]` option in select box.

maxiloc: got it makes sense

PR #605, opened 2016-09-30
PHP 5.3 compatibility fix

moloughlin:
This one was missed when restoring PHP 5.3 compatibility (causes a fatal error in the Magento configuration area)


JanPetr:
Great catch! Thanks a lot for contributing @moloughlin 🎉

PR #607, opened 2016-09-30
Fix to show correct bundle product prices

royduin:
When a bundle product does have a special price the old price isn't shown. This is because the `price_type` attribute isn't selected, see: https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Bundle/Model/Product/Price.php#L69-L76


JanPetr:
Thanks a lot for contributing @royduin! Really appreciated 🎉

royduin:
Thanks for merging 👍

PR #613, opened 2016-10-06
Add swedish translation

mikaelmattsson:


JanPetr:
Hello @mikaelmattsson,
I'm sorry for my late reply.
I'm not sure we want other translation files directly in the extension as it's hard to maintain (none of us here speaks Swedish) and it's not essential file for most of the users.

Do you have any limitation for which you need the translation file to be part of the extension? Let me know.

mikaelmattsson:
OK. Even if it is not maintained I think that it is better to have partial translation than none. And hopefully the community can patch the translation if needed.
I think it is rare that module translations are maintain by the module developers/team.

JanPetr:
OK, makes sense. Let's merge and see how it develops ;)
Thanks a lot for contributions.

PR #623, opened 2016-10-20
Improved removing of "No" values

### LINKED ISSUE #616: Index empty field values as empty rather than "No"
I'm using algolia 1.7.0. 

When algolia inserts _dropdown_ fields into the algolia index, if the field is blank, it enters a value of "No" - Instead, I think it should index a blank or null. 

Here's the use case where I'm running into a problem. 

I've added a field to my search results to show the manufacturer. So my newer template code looks like:

```
<div class="result-thumbnail">
    {{#image_url}}<img src="{{{ image_url }}}"/>{{/image_url}}
    {{^image_url}}<span class="no-image"></span>{{/image_url}}
</div>
{{#_highlightResult.manufacturer}}
    <div class="manufacturer">{{{_highlightResult.manufacturer.value}}}</div>
{{/_highlightResult.manufacturer}}
<div class="name">{{{_highlightResult.name.value}}}</div>
```

About 95% of my results have manufacturers, so it's 95% fine. But for the handful of products, there's no manufacturer, and in this field I get a "No" result - where instead I get a blank line.

So I'm seeing results like:

_[image]
No
White Socks_

and

_[image]
Levi's
Blue Jeans_

It looks like there's no conditional other than if-exists / if-doesn't-exist option in the hogan.js/mustache.js

Suggestions?

JanPetr:
Fixes #616

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
@@ -25,6 +25,8 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         'special_to_date',
     );
 
+    private $noAttributes = array();
+
     protected function getIndexNameSuffix()
     {
         return '_products';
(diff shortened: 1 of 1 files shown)


maxiloc:
👍

PR #626, opened 2016-10-21
Enable selecting autocomplete menu items

### LINKED ISSUE #625: Keyboard navigation
On the auto-complete search, you can use the down key, and then use the keyboard to navigate through the results.  It doesn't appear that you can navigate to the product page without clicking that result, though.

Is it possible to add <enter> to navigate to that result?

JanPetr:
Fixes #625


maxiloc:
👍

cvocke:
@JanPetr I was looking to fix this issue on the magento 2 plugin and found this. I applied the same fix to the magento 2 module and it worked great there as well

JanPetr:
Thanks @cvocke! 💯 
I'll port the change to M2 extension as well

PR #632, opened 2016-10-24
Ability to select sorts and ranking attributes from all products' attributes …

JanPetr:
… and not only those which are selected as indexable.

Selected attributes are indexed automatically with no need to index them explicitly.

cc @kokliKo


kokliKo:
\o/

maxiloc:
👍

PR #633, opened 2016-10-25
Fix reindexation when using full reindex of all ("catalogsearch_fulltext")

JanPetr:
Now it correctly detects enabled queue and do not perform delete operations.


maxiloc:
What was the issue ?

JanPetr:
@maxiloc: 
Issue is when you are reindexing with command: 
```sh
$ run php shell/indexer.php -- -reindex catalogsearch_fulltext
```
it'll reindex all indexers including Algolia's product indexer. But it bypasses indexer class, so it doesn't take indexing queue into account and trigger "regular" indexing without TMP indices.

Issue #631, opened 2016-10-24
Implement extension/hook/filter methods into autocomplete.js and instantsearch.js files


JanPetr:
The method(s) will accept different essential variables for autocomplete/instantsearch and allow it's users to modify them without need to directly edit the extension's files.

JanPetr:
will be part of the next release

PR #642, opened 2016-10-27
Add JS hook methods

JanPetr:
Implements #631


royduin:
I'm looking forward to these JS hooks!

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -257,6 +257,9 @@ $algoliaJsConfig = array(
 
 <script>
     window.algoliaConfig = <?php echo json_encode($algoliaJsConfig); ?>;
+    if (typeof algoliaHookAfterConfigLoad == 'function') {
+        window.algoliaConfig = algoliaHookAfterConfigLoad(algoliaConfig);
```
rayrutjes: Given the configuration is generated on the backend, I'm afraid it can be hard to debug if at a given point a user customizes in the backend and in the frontend with a hook.

js/algoliasearch/autocomplete.js
```diff
@@ -78,6 +78,10 @@ document.addEventListener("DOMContentLoaded", function(event) {
 				options.templates.footer = '<div class="footer_algolia"><span>' +algoliaConfig.translations.searchBy + '</span> <a href="https://www.algolia.com/?utm_source=magento&utm_medium=link&utm_campaign=magento_autocompletion_menu" target="_blank"><img src="' +algoliaConfig.urls.logo + '" /></a></div>';
 			}
 			
+			if (typeof algoliaHookBeforeAutocompleteStart == 'function') {
+				[sources, options] = algoliaHookBeforeAutocompleteStart(sources, options);
```
rayrutjes: I believe array destructuring is ES6 syntax.

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
JanPetr: Right now, we don't have a backend hook to modify generated configuration. So there'd be only one place to modify it - frontend.
Yes, having those hooks in general makes it harder to debug customers' issues, but from my experience those customers who are using hooks know what they are doing and do not have that kind of issues.

js/algoliasearch/autocomplete.js
JanPetr: Yes, you're right. Good catch, thanks. Will fix it.

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
rayrutjes: Makes sense 👍 💯

maxiloc:
👍  We'll just need the documentation along with that 😄

PR #650, opened 2016-10-31
Fix replace categories when categories are not set as attribute for f…

JanPetr:
…aceting. This sets categories attribute as attribute for faceting automatically.

### DIFF
--- a/app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
+++ b/app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
@@ -40,6 +40,7 @@ $query = '';
 $refinementKey = '';
 $refinementValue = '';
 $path = '';
+$level = '';
 
 $title = '';
 $description = '';
@@ -55,12 +56,17 @@ if ($config->isInstantEnabled() && $config->replaceCategories() && $app->getRequ
     if ($category && $category->getDisplayMode() !== 'PAGE') {
         $category->getUrlInstance()->setStore($storeId);
 
+        $level = -1;
         foreach ($category->getPathIds() as $treeCategoryId) {
             if ($path != '') {
                 $path .= ' /// ';
             }
 
             $path .= $productHelper->getCategoryName($treeCategoryId, $storeId);
+
+            if ($path) {
+                $level++;
+            }
         }
 
         $indexName = $productHelper->getIndexName($storeId);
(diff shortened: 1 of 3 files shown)


dmytro-butko:
thanks, 
I've
 - added 'categories' to Facets config section
 - upgraded to 1.7.2
And it works!

PS upgrading to 1.7.2 will make ones credentials incorrect (on any config change i.e. save?) as it seems:

-                            <frontend_type>password</frontend_type>
+                            <frontend_type>obscure</frontend_type>
+                            <backend_model>adminhtml/system_config_backend_encrypted</backend_model>

So integration will fail with ERROR Credentials not configured correctly unless you manually re-fill config details in Magento.
Was unexpected but easy to find in logs.

JanPetr:
Thanks for info @dmytro-butko,
I'm glad you made it work!

I added the not about Admin API key to release notes - https://github.com/algolia/algoliasearch-magento/releases/tag/1.7.2

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -119,15 +130,17 @@ if ($config->isInstantEnabled()) {
         $isSearchPage = true;
     }
 
+    foreach ($facets as $facet) {
```
maxiloc: What do you think about something like that:

```
array_in('categories', array_map(function ($facet) {
   return ($facet['attribute']);
}, $facets);
```

js/algoliasearch/instantsearch.js
```diff
@@ -75,6 +75,13 @@ document.addEventListener("DOMContentLoaded", function (event) {
 				if (algoliaConfig.request.refinementKey.length > 0) {
 					data.helper.toggleRefine(algoliaConfig.request.refinementKey, algoliaConfig.request.refinementValue);
 				}
+				
+				if (algoliaConfig.areCategoriesInFacets === false && algoliaConfig.request.path.length > 0) {
```
maxiloc: Not sure to get this part

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
JanPetr: Not sure about it as it's harder to read and I had to think for a while what your code does.
And it would be less efficient as it'd always iterate through all `$facets` and then again until it finds the correct attribute.

js/algoliasearch/instantsearch.js
JanPetr: I need to set the refinement only when categories widget does not exists and the user is on category page. Otherwise it breaks down categories hierarchical widget.

PR #653, opened 2016-11-01
Loading stockItems when required within Algolia_Algoliasearch_Helper_Entity_Producthelper::getObject

cmacdonald-au:
We have an external process that updates data directly to DB, thus doesn't trigger the usual indexing events.

To make sure updated products make it into the index, we're using the following as necessary;
`Mage::getModel('algoliasearch/indexer_algolia')->reindexSpecificProducts($productId);`

However that caused issues within the producthelper, specifically relating to it assuming that it will always have the stock item data inside the collection.

I've added the check & fetch condition for main and sub_products within `Algolia_Algoliasearch_Helper_Entity_Producthelper::getObject()`

Tested against Magento 1.7 specifically, (running in production), should work for all recent versions.

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
@@ -843,6 +843,11 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
         if (false === isset($defaultData['in_stock'])) {
             $stockItem = $product->getStockItem();
 
+            // getStockItem isn't always available
+            if (is_object($stockItem) === false) {
+                $stockItem = Mage::getModel('cataloginventory/stock_item')->loadByProduct($product);
+            }
+
             $customData['in_stock'] = (int) $stockItem->getIsInStock();
         }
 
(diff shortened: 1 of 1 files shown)


JanPetr:
Thanks for contributing @cmacdonald-au 🎉 Makes sense :)
Will test and if everything goes alright, I'll be happy to merge 💯

JanPetr:
Hello @cmacdonald-au,
I finally had time to review. In which cases it happens that stock item is not present? I cannot reproduce with current code.
You are running the code from the top, so it should be ok.

Anyway, I guess the better place to fix the issue will be directly in creating the collection to fetch stock data. What do you think? Can you update the PR?

cmacdonald-au:
True enough, will amend.

If it helps - we only saw it in the following (shortened) example (shell script - outside of the standard execution order);
```
$products = Mage::helper('eccbundles')->getAssociatedProductIdList($product);
Mage::log(__FILE__.'() Products ['.var_export($products, true).']');
$haveUpdates = Mage::helper('retailexpress')->syncProductStockById($products);

if ($haveUpdates) {
    Mage::log(__FILE__.'() Updating M2E model data');

    $model = Mage::getModel('M2ePro/PublicServices_Product_SqlChange');
    foreach ($products as $pid) {
    	$model->markQtyWasChanged($pid);
    }
    $model->applyChanges();

    Mage::log(__FILE__.'() Doing algolia');
    Mage::getModel('algoliasearch/indexer_algolia')->reindexSpecificProducts($product->getId());

    Mage::log(__FILE__.'() All done');
}
```

RetailExpress is "a bit invasive", so I'm guessing this is more of a "third party extension invalidates standard Mage model assumptions" than a straight vanilla issue.

JanPetr:
Hello, since I didn't hear from you and I'm still not able to reproduce, I'm closing this PR.
If you still experience the issue, feel free to re-open.

Thank you for your contribution :100:

cmacdonald-au:
Ah, sorry! We ended up resolving this one upstream, (outside of this extension). I should've closed this much earlier :)

PR #666, opened 2016-11-11
Added new event dispatching

firewizard:
Dispatching event algolia_rebuild_store_product_index_page_collection_load_before for each page when indexing products, so the collection can be manipulated before loading it


JanPetr:
Hey @firewizard,
sorry for the late answer! Thanks for contribution. Will test it and merge after 👍

JanPetr:
Hey @firewizard,
I finally found time to review and test.
Could you rename the event so there is no conflict between [Producthelper.php#L199](https://github.com/algolia/algoliasearch-magento/blob/develop/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php#L199) and this change?
And could you pass `$storeId` variable to the event as well?

Thanks!

firewizard:
Hi @JanPetr ,

Not sure I'm following. That event has been added after I submitted the PR, but that's not the real issue. The naming of my event was based on `algolia_[method_name]_[event]`. the naming in the producthelper class doesn't really make sense, not to mention you're triggering two events with the same data which also doesn't make sense to me.

I can pass the storeId param, sure, and I can change the name too - just tell me what name to give it.

JanPetr:
You are right, two events are called before we fixed naming, but didn't want to break backwards compatibility. That's why two events are called. The first one has comment that it's there only because of BC.

Can you name it `algolia_before_products_collection_load`? That'd describe it the best. WDYT?

firewizard:
sure, i'll rename it tomorrow and then push the changes.

JanPetr:
Thanks @firewizard, I'm looking forward

firewizard:
@JanPetr I've renamed the event and also added $storeId. Seems the upon sync'ing my fork there are more changes in the PR...

JanPetr:
Thanks @firewizard!

PR #702, opened 2016-12-13
Ability to hide categories & products from dropdown menu

### LINKED ISSUE #701: Ability to hide the Categories from autocomplete menu
From what I understand from [that line](https://github.com/algolia/algoliasearch-magento/blob/master/js/algoliasearch/autocomplete.js#L30) you cannot remove the display of categories.

I think we should have the same logic than for queries -> if 0, then it's hidden.

redox:
Something like that @JanPetr?

Fix #701


JanPetr:
Thanks for bootstrapping if @redox 👍 

I changed comments in configuration and simplified it a bit. Can you review?

PR #708, opened 2016-12-26
Fixes the bug when searchableAttributes setting is returned from API …

### LINKED ISSUE #700: An error occurred while saving this configuration: You cannot specify "attributesToIndex" and "searchableAttributes" at the same time. "attributesToIndex" is deprecated you should only use "searchableAttributes"
/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php needs to be updated to use searchableAttributes instead of attributesToIndex

Currently the module only uses attributesToIndex, but app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php:mergeSettings will fetch searchableAttributes from Algolia settings and try to merge them together. This results in the final API request object containing both attributesToIndex and searchableAttributes.

JanPetr:
…and then both searchableAttributes and attributesToIndex are set to index which results in error

Fixes #700

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
@@ -84,6 +84,16 @@ class Algolia_Algoliasearch_Helper_Algoliahelper extends Mage_Core_Helper_Abstra
         } catch (\Exception $e) {
         }
 
+        if (isset($settings['attributesToIndex'])) {
+            $settings['searchableAttributes'] = $settings['attributesToIndex'];
+            unset($settings['attributesToIndex']);
+        }
+
+        if (isset($onlineSettings['attributesToIndex'])) {
+            $onlineSettings['searchableAttributes'] = $onlineSettings['attributesToIndex'];
+            unset($onlineSettings['attributesToIndex']);
+        }
+
         $removes = array('slaves');
 
         foreach ($removes as $remove) {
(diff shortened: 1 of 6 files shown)


app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
```diff
@@ -84,6 +84,11 @@ public function mergeSettings($index_name, $settings)
         } catch (\Exception $e) {
         }
 
+        if (isset($onlineSettings['searchableAttributes']) && isset($settings['attributesToIndex'])) {
+            $settings['searchableAttributes'] = $settings['attributesToIndex'];
+            unset($settings['attributesToIndex']);
+        }
+
```
maxiloc: I would go even further than that and replace by searchAttributes even when not the both of them are set

JanPetr: Updated, can you check?

app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
```diff
@@ -84,6 +84,16 @@ public function mergeSettings($index_name, $settings)
...
+        if (isset($onlineSettings['attributesToIndex'])) {
+            $onlineSettings['searchableAttributes'] = $onlineSettings['attributesToIndex'];
+            unset($onlineSettings['attributesToIndex']);
+        }
+
         $removes = array('slaves');
```
maxiloc: Yes better, it would make sense to rename it when we set it as well (https://github.com/algolia/algoliasearch-magento/blob/037bc15692035dbbc9c4d4cb67705f4455238478/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php#L269), wdyt ?

JanPetr: Yes, let's do it now. Next release we gonna have events there and it'll cause a BC break. 
Good catch, thanks.

PR #709, opened 2016-12-26
Fix auto complete menu UX on mobile. Now it's impossible to hide keyb…

### LINKED ISSUE #703: Default autocomplete menu behavior on mobile is not perfect
Using https://magento.algolia.com/, on iPhone. When using the autocomplete menu, if I want to hide the keyboard after typing "a" to see all results and scroll, it will hide all results.

Could we maybe on the mobile experience put `debug: true` on the autocomplete menu so that on mobile it does not disappears on blur?

It was also sent as feedback on algolia support and after trying it I thought that yes it could be better.

JanPetr:
…oard and browse auto complete menu results as the menu disappears on blur. Setting debug on true on mobile devices fixes the issue.

Fixes #703

### DIFF
--- a/js/algoliasearch/autocomplete.js
+++ b/js/algoliasearch/autocomplete.js
@@ -74,6 +74,11 @@ document.addEventListener("DOMContentLoaded", function(event) {
 				debug: false
 			};
 			
+			if (isMobile() === true) {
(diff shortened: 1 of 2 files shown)


maxiloc:
good for me

Issue #729, opened 2017-01-16
Automatically add "updateAt" timestamp equals to NOW() to a record pushed to Algolia


JanPetr:

JanPetr:
Implemented in #730, it'll be part of the next release

PR #730, opened 2017-01-17
Add attribute "algoliaLastUpdateCET" to all objects indexed via the e…

JanPetr:
…xtension. It contains the exact CET when the record was update for the last time. It helps to debug indexing issues.

Implements #729

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
@@ -109,52 +109,11 @@ class Algolia_Algoliasearch_Helper_Algoliahelper extends Mage_Core_Helper_Abstra
         return $onlineSettings;
     }
 
-    public function handleTooBigRecords(&$objects, $index_name)
+    public function addObjects($objects, $indexName)
     {
-        $long_attributes = array('description', 'short_description', 'meta_description', 'content');
+        $this->prepareRecords($objects, $indexName);
 
-        $good_size = true;
-
-        $ids = array();
-
-        foreach ($objects as $key => &$object) {
-            $size = mb_strlen(json_encode($object));
-
-            if ($size > 20000) {
-                $good_size = false;
-
-                foreach ($long_attributes as $attribute) {
-                    if (isset($object[$attribute])) {
-                        unset($object[$attribute]);
-                        $ids[$index_name.' objectID('.$object['objectID'].')'] = true;
-                    }
-                }
-
-                $size = mb_strlen(json_encode($object));
-
-                if ($size > 20000) {
-                    unset($objects[$key]);
-                }
-            }
-        }
-
-        if (count($objects) <= 0) {
-            return;
-        }
-
-        if ($good_size === false) {
-            /** @var Mage_Adminhtml_Model_Session $session */
-            $session = Mage::getSingleton('adminhtml/session');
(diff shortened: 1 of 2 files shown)


maxiloc:
I would actually do it here https://github.com/algolia/algoliasearch-magento/blob/master/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php#L155-L163

JanPetr:
@maxiloc I was thinking about doing it there, but I didn't what to iterate again over all the objects.

app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
```diff
@@ -109,52 +109,37 @@ public function mergeSettings($index_name, $settings)
...
+    public function addObjects($objects, $indexName)
     {
-        $long_attributes = array('description', 'short_description', 'meta_description', 'content');
+        $currentCET = new DateTime('now', new DateTimeZone('Europe/Paris'));
+        $currentCET = $currentCET->format('Y-m-d H:i:s');
+        
+        $modifiedIds = array();
 
-        $good_size = true;
+        foreach ($objects as $key => $object) {
```
maxiloc: I would do this loop in another function

JanPetr: Why?
The reason the loop is here I don't want to mix "add time" and "handle big records" features together in any other function. The only thing those "features" have in common is the "addObjects" method.

maxiloc: To keep it very simple like other method of the file:
https://github.com/algolia/algoliasearch-magento/blob/221492f938275674b3b0f4865a26dc1f8ece2ffd/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php#L34-L76
that are just a bridge where we can do stuff

I tend to prefer seeing:
```
  if ($this->config->isPartialUpdateEnabled()) {
            $index->partialUpdateObjects($objects);
        } else {
            $index->addObjects($objects);
        }
```
very quickly and not at the end of a long function

The method name can be generic like `alterRecords` or `modifyRecords` and inside you can have two different submethods

JanPetr: Makes sense, stay tuned

JanPetr: @maxiloc Extracted, can you review?

app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php
```diff
@@ -109,52 +109,37 @@ public function mergeSettings($index_name, $settings)
...
         }
 
-        if ($good_size === false) {
+        if (!empty($modifiedIds)) {
             /** @var Mage_Adminhtml_Model_Session $session */
             $session = Mage::getSingleton('adminhtml/session');
-            $session->addError('Algolia reindexing : You have some records ('.implode(',',
-                        array_keys($ids)).') that are too big. They have either been truncated or skipped');
+            $session->addError('Algolia reindexing : You have some records ('.implode(',', $modifiedIds).') that are too big. They have either been truncated or skipped');
         }
```
maxiloc: I think you are missing the return

Issue #595, opened 2016-09-22
Indexing Only Products which Have Visibility-Search = 1


lemming76:
Hi,

Can anyone tell me if this is possible and if not can it be added?

I have say 200 grouped products and 10,000 simple products active in the magento database.  Of the 10,000 simple products, 6,000 of these may be sub-products of the 200 grouped products.

All I want indexing in Algolia are the 200 grouped products and 4000 simple products which are not part of the grouped products.  In my system, all simple products which are part of grouped products only have visibility set as catalog, not search.

At the moment I can see these being indexed in Algolia with the attribute Visibility-Search: 0

Is there anyway that only products in the database which have Visibility-Search: 1 are uploaded?  It just seems to make sense that if you have said that they shouldn't be found in the search, that they do not get indexed by Algolia.

Thanks in advance for any help.

JanPetr:
Hello @lemming76,

the extension indexes products visible only in catalog, because you may want to use Algolia for category listing, not only search.
When you are searching on your website, only products with `visibility_search: 1` are searched, when you are browsing your website and uses Algolia for category listing, only products with `visbility_catalog: 1` are fetched.

Right now, there is no UI option how to prevent products with only catalog visibility from indexing. You can update your code [here](https://github.com/algolia/algoliasearch-magento/blob/master/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php#L133) and [here](https://github.com/algolia/algoliasearch-magento/blob/master/app/code/community/Algolia/Algoliasearch/Helper/Data.php#L463-L467) to not index them.

It'd be probably nice to have such a feature built-in. I'll keep the issue open and we'll target it in upcoming versions.

Thanks for bringing it up.

lemming76:
Hi @JanPetr 

Thanks for such a quick response.  I appreciate it.  It would be great if such a feature could be included in the UI in a future version.  Thanks for considering the request.

Cheers

davids101:
would like to see this as well

JanPetr:
Implemented in #731. It'll be part of the next release.

PR #731, opened 2017-01-18
Add option to index only products with specific visibility

JanPetr:
Implements #595


maxiloc:
Why do we need an option for that ?
We probably can do it without. The reason we index everything was mainly debugging, right now this part of the code is pretty stable and we can probably remove it

PR #736, opened 2017-01-21
Close the DB connection when the queue runner is done

JanPetr:


maxiloc:
Isn't supposed to be done by Magento

JanPetr:
Should be, but we had issue from a customer about alive connections from QueueRunner.
So I wouldn't rely on it anymore and close it manually here.

maxiloc:
makes sense but I am a bit worried about magento needing the connection somehow and failing (not in our code) because there is no connection anymore

JanPetr:
I tested it and didn't find any issue with that. When merged, it's going to live in `develop` branch for a while and we will see if it's causing issues on different instances.

maxiloc:
👍

PR #742, opened 2017-01-31
Bugfix, no search results when Algolia is disabled

royduin:
When Algolia is disabled with `algoliasearch/credentials/enable_backend` for a store it's not returning any search results.


JanPetr:
Thanks @royduin, great catch! 👍

PR #754, opened 2017-02-10
Add analytics feature

JanPetr:


maxiloc:
Just to be sure, we are out of the box we GA if you enable it ?
And universal analytics handle all of them ?

rayrutjes:
Yes @maxiloc : https://github.com/algolia/algoliasearch-magento/pull/754/files#diff-146d9fc56a2377974bb77666381a1535R408

maxiloc:
Yep I saw that 😄  but is all analytics except google are using a `ga` variable ?

JanPetr:
You can implement GA into a website in two ways. In [legacy way](https://developers.google.com/analytics/devguides/collection/gajs/methods/gaJSApi_gaq) (using `_gaq.push()`) or like [Universal Analytics](https://developers.google.com/analytics/devguides/collection/analyticsjs/pages) (`ga` object).
Magento out-of-the-box supports both:
screenshot 2017-02-10 16 28 36

That's why I test both and use the right code for the one enabled.

If the owner of the website uses something different or want to use it in any other way, he can [define his own push function ](https://github.com/algolia/algoliasearch-magento/pull/754/files#diff-146d9fc56a2377974bb77666381a1535R403).

PR #757, opened 2017-02-15
Queue merge optimalization and delete operations optimalization

JanPetr:


app/code/community/Algolia/Algoliasearch/Model/Queue.php
```diff
@@ -16,7 +16,16 @@ class Algolia_Algoliasearch_Model_Queue
     /** @var Algolia_Algoliasearch_Helper_Logger */
     protected $logger;
 
-    protected $by_page;
+    protected $maxSingleJobDataSize;
+
+    private $staticJobMethods = array(
```
maxiloc: Why do we call that one static ?

app/code/community/Algolia/Algoliasearch/Model/Queue.php
```diff
@@ -59,184 +68,274 @@ public function runCron()
...
+        $limit = $maxJobs = ($maxJobs === -1) ? $this->config->getNumberOfJobToRun() : $maxJobs;
+        $offset = 0;
+
+        $actualBatchSize = 0;
+        $maxSingleJobDataSize = $this->config->getNumberOfElementByPage() * $limit;
+
+        try {
+            $this->db->beginTransaction();
+
+            while (true) {
```
maxiloc: We can probably avoid the while true

app/code/community/Algolia/Algoliasearch/Model/Queue.php
JanPetr: Because those jobs cannot be reordered or merged

app/code/community/Algolia/Algoliasearch/Model/Queue.php
JanPetr: Fixed

Mobecls:
@JanPetr  and how did you optimized deletion? We still have large amount of deletions (half of all of our operations) and most of our products are configurables...

Mobecls:
@JanPetr can we verify if the simple product, that has "not visible individually" is a part of configurable product. And if so, do not send "delete" operation?

JanPetr:
Hey @Mobecls,
do you have enabled indexing queue? The optimalization works best with queue enabled, when products are indexed by batches.

You can do that by modifying the extension, but I won't put it directly to the extension as there is a chance you will have indexed wrong data.

Mobecls:
@JanPetr  yes, we have queue enabled.

The logic is that not visible individually, which is part of configurable - should not be shown and won't be shown evern - nor on website, no on algolia. So there is no need to send delete's....

JanPetr:
The issue is it might be indexed before becoming part of the configurable product.

Mobecls:
I was thinking of yes/no setting

PR #772, opened 2017-02-24
Define clearfix CSS class

### LINKED ISSUE #771: missing clearfix class
`<div class="clearfix"></div>`

This snippet appears in the wrapper.phtml file, but I don't see it defined anywhere in the algoliasearch.css file.  I'm assuming this is to cover the height of the floated elements above it, and it is probably a dependency on bootstrap?  Our site, however, doesn't contain bootstrap at all, so it's not formatting perfectly.  

I'd recommend(assuming I'm correct and not missing something, which is always possible) that the documentation either reflects a dependency on bootstrap, or the class be renamed algolia-clearfix and defined in the included css.

JanPetr:
Fixes #771

### DIFF
--- a/skin/frontend/base/default/algoliasearch/algoliasearch.css
+++ b/skin/frontend/base/default/algoliasearch/algoliasearch.css
@@ -1212,3 +1212,9 @@ a.ais-current-refined-values--link:hover
 #algolia-right-container .infos.hidden {
 	display: none;
 }
+
+.algolia-clearfix:after {
+	content:"";
+	display:block;
+	clear:both;
+}
--- a/app/design/frontend/base/default/template/algoliasearch/autocomplete/attribute.phtml
+++ b/app/design/frontend/base/default/template/algoliasearch/autocomplete/attribute.phtml
@@ -4,6 +4,5 @@
         <div class="info-without-thumb">
             {{{_highlightResult.value.value}}}
         </div>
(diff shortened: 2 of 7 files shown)


rayrutjes:
The ":before" is not required for the clearfix to work.
Also you can add the classes to the parent container instead of adding empty divs.
Regarding the zoom rule, it is there to support IE6/7, not longer needed.
http://cssmojo.com/the-very-latest-clearfix-reloaded/

JanPetr:
Good catch, thanks @rayrutjes. Will change

skin/frontend/base/default/algoliasearch/algoliasearch.css
```diff
@@ -1211,4 +1211,10 @@ a.ais-current-refined-values--link:hover
 
 #algolia-right-container .infos.hidden {
 	display: none;
+}
+
+.algolia-clearfix:after {
+	content:"";
+	display:block;
```
rayrutjes: On the shared link, they use `display:table`, it probably addresses some edge case so I think it is better to use it instead of `block`.

JanPetr: Actually in the link he's favoring `block`:
screenshot 2017-02-24 09 55 14

rayrutjes:
Correct, I overlooked that, makes sense, thank you for spotting that!

Issue #762, opened 2017-02-20
Create debug / support script


JanPetr:
The script dumps all config values and other debug-related data to a single log file, which can be sent to Algolia support in terms of issues with the extension.

JanPetr:
Implemented in 1.9.0

PR #777, opened 2017-02-27
Implement debug dump script

JanPetr:
Implements #762


aude:
Just a quick note, this could be put at eg. `shell/algolia_dump_config.php` or `shell/algolia/dump_config.php`. CLI scripts usually go in `shell` in the Magento 1 structure.

PR #789, opened 2017-03-14
Add translation support / initialize translator

michaelscheel:


JanPetr:
Hello @michaelscheel,
thanks for the contribution!

Would you mind elaborate a bit on why this feature is needed and what it brings to the extension? Thanks!

michaelscheel:
Hi @JanPetr, sorry, sure. When using event algolia_product_index_before to add custom product data it's not possible to use integrated translation functionality (__()), because the translator is not initialized yet. It would be possible that the event observer does that, but if you have multiple observers in different modules it's not a clean approach and you're doing it multiple times.

JanPetr:
Makes sense @michaelscheel, thanks.
Would you mind putting a short comment to the line so it's seen why it's there? :) Thanks!

michaelscheel:
Sehr geehrte Damen und Herren,

ich befinden mich bis einschließlich 20.03.2017 nicht im Büro. In dringenden Fälle wenden Sie sich bitte an meinen Kollegen Tobias Völker voelker@virtualways.de.

Mit freundlichen Grüßen 

Michael Scheel

PR #792, opened 2017-03-16
Disabling Search and Indexing options are independant

julienbourdeau:
* The option called _Enable Extension_ was renamed to _Enable Indexing_, the good news is that the code was actually written this way.
* `isEnabledFrontEnd()` method only relies on Frontend option
* In config admin page, _Enable frontend_ and _Enable Logging_ don't rely on _Enable Indexing_ anymore.

See: https://discourse.algolia.com/t/using-main-web-search-on-all-store-locations/697

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Config.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Config.php
@@ -145,8 +145,7 @@ class Algolia_Algoliasearch_Helper_Config extends Mage_Core_Helper_Abstract
     public function isEnabledFrontEnd($storeId = null)
     {
         // Frontend = Backend + Frontent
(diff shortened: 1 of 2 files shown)


app/code/community/Algolia/Algoliasearch/Helper/Config.php
```diff
@@ -145,8 +145,7 @@ public function displaySuggestionsCategories($storeId = null)
     public function isEnabledFrontEnd($storeId = null)
     {
         // Frontend = Backend + Frontent
```
JanPetr: This comment is not needed anymore

julienbourdeau:
I can't change the origin branch on opened pull request, so I'll open another one taking into consideration your comments 👍

PR #793, opened 2017-03-17
Make indexing and search options independant

julienbourdeau:
This means that if you turn off indexing, for whatever reasons, your can still search them on your frontend.

I think it's better to not delete the indices when the indexing option is turned off because you might need to turn it off temporarily (maintenance, debugging or so) but you still want your frontend to have the Algolia search bar.

**Related:**
* Previous PR #792
* https://discourse.algolia.com/t/using-main-web-search-on-all-store-locations/697

## Notes
 * I added a filter option to the `runTests.sh` script so we can choose which tests to run while developing 🎉 
 * Default configuration is now re-applied before each tets

### DIFF
--- a/tests/AbstractIndexingTestCase.php
+++ b/tests/AbstractIndexingTestCase.php
@@ -7,9 +7,15 @@ class AbstractIndexingTestCase extends TestCase
     /** @var Algolia_Algoliasearch_Helper_Algoliahelper */
     protected $algoliaHelper;
 
+    protected $indexPrefix;
+
     public function setUp()
     {
         $this->algoliaHelper = Mage::helper('algoliasearch/algoliahelper');
+
+        /** @var Algolia_Algoliasearch_Helper_Config $config */
+        $config = Mage::helper('algoliasearch/config');
+        $this->indexPrefix = $config->getIndexPrefix();
 
         setConfig('algoliasearch/queue/active', '0');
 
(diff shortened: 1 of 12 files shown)


tests/QueueTest.php
```diff
@@ -377,10 +369,6 @@ public function testGetJobs()
 
     public function testHugeJob()
     {
-        // Default value - maxBatchSize = 1000
-        setConfig('algoliasearch/queue/number_of_job_to_run', 10);
```
JanPetr: Why did you remove those?

tests/AbstractIndexingTestCase.php
```diff
@@ -7,30 +7,45 @@ class AbstractIndexingTestCase extends TestCase
...
+    protected $indexPrefix;
+
+    /**
+     * @var array Default value of general store settings (not related to algolia's extention)
+     */
+    protected $defaultConfig = array(
+        'cataloginventory/options/show_out_of_stock' => '0',
+    );
+
+    const DEFAULT_PRODUCT_COUNT = 86;
```
JanPetr: Does it make sense to have it just for products?

tests/ProductsIndexingTest.php
```diff
@@ -91,4 +78,24 @@ public function testDefaultIndexableAttributes()
...
+
+        $productIndexer = new Algolia_Algoliasearch_Model_Indexer_Algolia();
+        $this->processTest($productIndexer, 'products', self::DEFAULT_PRODUCT_COUNT);
+    }
+
+    public function testProductAreSearchableIfIndexingIsDisabled()
+    {
+        setConfig('algoliasearch/credentials/enable_backend', '0');
+
+        $this->processQueryOneProduct();
```
JanPetr: I don't get what this test do? It set's config, but don't index anything and just query one product?

JanPetr: Oh, my bad. I just read the name of the test :)

JanPetr: Why did you extract it to a method anyway, when you don't reuse it? Or do you plan to?

tests/AbstractIndexingTestCase.php
julienbourdeau: We could add one constant for each but the other ones are only used once. They are specific for the test, this one is what you get with the default config.

tests/QueueTest.php
julienbourdeau: Because they settings the default value, which are set in `setUp()` method. I might have missed something though ^^

tests/ProductsIndexingTest.php
julienbourdeau: When I wrote it I thought I would reuse it but in the end I only wrote one test. Should I move it back in `ProductsIndexingTest`?

tests/AbstractIndexingTestCase.php
JanPetr: I'd move the constant at least to ProductIndexingTest.php as it's relevant only to products and not other children of the class.

tests/ProductsIndexingTest.php
JanPetr: Yes please :)

tests/QueueTest.php
```diff
@@ -1,8 +1,6 @@
 <?php
 
-use PHPUnit\Framework\TestCase;
-
-class QueueTest extends TestCase
+class QueueTest extends AbstractIndexingTestCase
```
JanPetr: Why this class inherits AbstractIndexingTestCase when it's not really testing indexing?
If it just need it's dependencies I think the better approach would be to create AbstractTestCase, put all common stuff (like $indexPrefix) there and remove those stuff form AbstractIndexingTestCase.
Then make AbstractIndexingTestCase inherit from AbstractTestCase.
So we avoid use less code inherited everywhere. WDYT?

tests/QueueTest.php
JanPetr: That's true, but I'd keep it here said explicitly, because the is dependent on this value. And in case you change default values in XML file, you might break the tests or make it false positive.

The same applies to testMaxSingleJobSize() method.

tests/bootstrap.php
```diff
@@ -29,10 +27,18 @@ function resetConfigs($configs = array())
...
 
+    $credentialsShortcuts = array(
+        'credentials/application_id', 'credentials/api_key', 'credentials/search_only_api_key', 'credentials/index_prefix'
+    );
+
     foreach ($xml->default->algoliasearch->children() as $section => $subsections) {
         foreach ($subsections as $subsectionName => $subsection) {
             $shortcut = $section.'/'.$subsectionName;
 
+            if (in_array($shortcut, $credentialsShortcuts)) {
```
JanPetr: 👍

tests/QueueTest.php
julienbourdeau: Definitely! I'll create `AbstractTestCase` class.

JanPetr:
👍  Feel free to merge! 💯  🎉

PR #794, opened 2017-03-17
Optionally install xdebug when running test

julienbourdeau:


JanPetr:
Good job @julienbourdeau! 👍

PR #796, opened 2017-03-20
Add infinite scroll feature

julienbourdeau:
If you turn the option on, you'll see a _load more_ button instead of the traditional pagination.

 - [x] Add configuration
 - [x] Add frontend feature
 - [x] Dig deeper url behavior

Any comments are welcome while I'm adding tests.

### Backend

configuration___system___magento_admin

### Frontend

men

### DIFF
--- a/app/design/frontend/base/default/template/algoliasearch/instantsearch/hit.phtml
+++ b/app/design/frontend/base/default/template/algoliasearch/instantsearch/hit.phtml
@@ -14,61 +14,61 @@ $price_key = $config->isCustomerGroupsEnabled(Mage::app()->getStore()->getStoreI
 
 <!-- Product hit template -->
 <script type="text/template" id="instant-hit-template">
-    {{#hits}}
-        <div class="col-md-4 col-sm-6">
-            <div class="result-wrapper">
-                <a href="{{url}}" class="result algolia-clearfix">
-                    <div class="result-content">
-                        <div class="result-thumbnail">
-                            {{#image_url}}<img src="{{{ image_url }}}"/>{{/image_url}}
-                            {{^image_url}}<span class="no-image"></span>{{/image_url}}
-                        </div>
-                        <div class="result-sub-content">
-                            <h3 class="result-title text-ellipsis">
-                                {{{ _highlightResult.name.value }}}
-                            </h3>
-                            <div class="ratings">
-                                <div class="ratings-wrapper">
-                                    <div class="ratings-sub-content">
-                                        <div class="rating-box">
-                                            <div class="rating" style="width:{{rating_summary}}%" width="148" height="148"></div>
-                                        </div>
-                                    </div>
-                                    <div class="price">
-                                        <div class="price-wrapper">
-                                            <div>
-                                                <span
-                                                    class="after_special {{#price<?php echo $price_key; ?>_original_formated}}promotion{{/price<?php echo $price_key; ?>_original_formated}}">
-                                                    {{price<?php echo $price_key; ?>_formated}}
-                                                </span>
 
-                                                {{#price<?php echo $price_key; ?>_original_formated}}
-                                                    <span class="before_special">
-                                                        {{price<?php echo $price_key; ?>_original_formated}}
-                                                    </span>
-                                                {{/price<?php echo $price_key; ?>_original_formated}}
-                                            </div>
+    <div class="col-md-4 col-sm-6">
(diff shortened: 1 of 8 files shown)


app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -193,6 +193,9 @@ $autocompleteApiKey = $algoliaHelper->generateSearchSecuredApiKey($config->getSe
 $algoliaJsConfig = array(
     'instant' => array(
         'enabled' => $config->isInstantEnabled(),
+        'infiniteScroll' => array(
+                'enabled' => $config->isInfiniteScrollEnabled(),
```
JanPetr: Does it have to be nested array? How about just `enabledInfiniteScroll`?

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -271,6 +274,7 @@ $algoliaJsConfig = array(
         'categories' => $this->__('Categories'),
         'products' => $this->__('Products'),
         'searchBy' => $this->__('Search by'),
+        'showMore' => $this->__('More products'),
```
JanPetr: I'd go for more descriptive "Show more products" than just "More products"

js/algoliasearch/instantsearch.js
```diff
@@ -159,27 +159,44 @@ document.addEventListener("DOMContentLoaded", function (event) {
...
-			})
-		);
+				})
+			);
+		} else {
+			search.addWidget(
+				algoliaBundle.instantsearch.widgets.hits({
+					container: '#instant-search-results-container',
+					templates: {
+						allItems: '{{#hits}}' + $('#instant-hit-template').html() + '{{/hits}}'
```
JanPetr: That we cannot do, people need to have option to update template without touching JS code.
And it's a huge BC break as current users may have custom template and this code would make it not work at all.

redox:
That's so cool @JanPetr @julienbourdeau 👍  Does it still store the `page` query parameter in the URL? What's the impact when you share the URL with someone while being already on page 3?

app/design/frontend/base/default/template/algoliasearch/instantsearch/hit.phtml
```diff
@@ -14,61 +14,61 @@ $price_key = $config->isCustomerGroupsEnabled(Mage::app()->getStore()->getStoreI
 
 <!-- Product hit template -->
 <script type="text/template" id="instant-hit-template">
-    {{#hits}}
```
JanPetr: Huge BC break, I wouldn't modify this template. I'd rather create a new template like `infinite-scroll-hit.phtml` and copy it.
It'll be harder for the maintenance, but better for BC POV, what do you think?

Or, you can create a template `simple-hit.phtml` and include it here in `hit.phtml`, so we would have code written only once with preserving BC. Hmm?

julienbourdeau:
@redox So the page parameter is ignored but it seems like other parameters are well. I'll have a deeper look. Thanks!

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
julienbourdeau: I had translations nested here before but I moved them. You're right, I'll remove one level.

julienbourdeau:
Ok so the page parameters is now removed from the URL when infinite scroll is enabled.

Thank you all for your comments! 👍

tataragne:
Hi, nice improvment ! Can you rebase this branch on 1.9.0 version ?

julienbourdeau:
@tataragne I just rebased it on top of the develop branch 👍

tataragne:
@julienbourdeau ty, @JanPetr any chance to get this in the next release ?

JanPetr:
@julienbourdeau I tested it and everything looks OK! Just one small thing - when I go to the end, there is still "Show more products" button displayed. Can you hide it on the last page?

julienbourdeau:
Oh yeah, I'll look into it!

julienbourdeau:
I added it. First, I starter something overly complicated with a js eventlistener but then I found a much better solution. ^^

magegirl:
Hello,
Is it possible that the scroll automatically starts without button "show more" ?

PR #798, opened 2017-03-23
Enable Rewrite for Algolia_Algoliasearch_Model_Resource_Engine

michaelscheel:
The engine shouldn't be loaded via direct initialization and should instead follow Magento best practice to allow rewrites by third party developers.

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Model/Indexer/Abstract.php
+++ b/app/code/community/Algolia/Algoliasearch/Model/Indexer/Abstract.php
@@ -11,7 +11,7 @@ abstract class Algolia_Algoliasearch_Model_Indexer_Abstract extends Mage_Index_M
     {
         parent::__construct();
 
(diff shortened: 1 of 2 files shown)


michaelscheel:
I guess it makes no sense to add things to the queue multiple times (e.g. when a product is saved multiple times within a few minutes and the queue is already full).

julienbourdeau:
Hi @michaelscheel,

Thanks for your contribution!

Would you mind cleaning your branch so you only have the first commit? The following commands should of the trick.

```
git reset --hard fac05593aba5c9ca870a6459d441ca22afe6bbc5
git push -f
```

michaelscheel:
Hi @julienbourdeau,

ok - it's done.

julienbourdeau:
I have run the tests locally and they pass 👍 

Thanks for your contribution!

PR #803, opened 2017-03-29
Introduce end2end tests for Magento

julienbourdeau:
There are still the tests running in docker container to fix

And I need to write the documentation.


dev/bin/test.sh
```diff
@@ -40,7 +40,7 @@ cd /var/www/htdocs/.modman/algoliasearch-magento
 if [ $FILTER ]; then
     vendor/bin/phpunit tests/backend --filter "$FILTER"
 else
-    vendor/bin/phpunit tests/backend
+    # vendor/bin/phpunit tests/backend
```
julienbourdeau: I assume this was for development purpose

tests/frontend/specs/autocomplete/autocomplete.js
```diff
@@ -6,6 +6,7 @@ describe('Test Autocomplete features', function () {
     });
 
     it('should show the algolia search form', function () {
+    	console.log(browser.url('/').getSource());
```
julienbourdeau: And here maybe, or it's intended to print this to help debug.

julienbourdeau: 👍

Issue #813, opened 2017-04-27
Weee is not indexed


trosyno:
Hello,

Weee is not added to the price when products are indexed. Is there a fix for that?

Thank you

JanPetr:
Hello @trosyno,
maybe a stupid question, but ... what is "Weeee"? Thanks!

trosyno:
Hello,

Weee is "Waste Electrical and Electronic Equipment Directive", a tax in Europe for electronic products. The module that handles it is Mage_Weee

Here's how it works:
![weee1](https://cloud.githubusercontent.com/assets/8102461/25619414/12703300-2f4b-11e7-8e16-5a81355eb80b.jpg)
![weee2](https://cloud.githubusercontent.com/assets/8102461/25619473/4a48f6e0-2f4b-11e7-9a85-fc7edd6e46d4.jpg)
![weee3](https://cloud.githubusercontent.com/assets/8102461/25619554/9dd89b26-2f4b-11e7-96d7-af4f766c887c.jpg)

The 0.25 tax will be included in the product final price

JanPetr:
Thanks for explanation. I will take a look and see what can be done.

trosyno:
Thank you. Our need is quite urgent so we may get to do it before you do. If so, the method we need to alter is handlePrice in Algolia_Algoliasearch_Helper_Entity_Producthelper, right?

JanPetr:
Hello @trosyno,

yes, you are completely right. If you'll manage to fix it, it'll be awesome if you could send a PR with the fix :) Thanks!

JanPetr:
Implemented in #829, it'll be part of the next release.

PR #829, opened 2017-05-15
Add weee tax to product price

trosyno:
Adds weee tax (European tax on electronic products) to the price
see #813


trosyno:
#829

JanPetr:
Thanks for the PR @trosyno!
Is the WEEE module always present in Magento instance? Is that something we can rely on? I'd put there a check if the WEEE is present and enabled.
Could you add it?

trosyno:
You're right! I've just added the check.

JanPetr:
Thanks @trosyno, everything seems to be working now! Thanks for your contribution 💯

Issue #831, opened 2017-05-17
Two queries are executed when visiting category page url without query parameters


Inwerpsel:
When instantsearch on category pages is enabled then two queries are executed.

Steps to reproduce:
- go to [https://magento.algolia.com/accessories/eyewear.html](https://magento.algolia.com/accessories/eyewear.html)
- open network pane, filter on XHR and refresh
- there are two almost identical requests for the search

This seems to be caused by [https://github.com/algolia/algoliasearch-magento/blob/master/js/algoliasearch/instantsearch.js#L461-L467](https://github.com/algolia/algoliasearch-magento/blob/master/js/algoliasearch/instantsearch.js#L461-L467): first search.start() is called then search.search() is called.

JanPetr:
Thanks for the report @Inwerpsel, I investigate and come back to you.

Inwerpsel:
Wouldn't it make more sense to use the url to know if any other category filter is applied instead of executing the search just to know this?

Probably it is possible to set the category parameter after the search object is instantiated with on https://github.com/algolia/algoliasearch-magento/blob/master/js/algoliasearch/instantsearch.js#L66 but before calling search,start().

JanPetr:
It'd be the best, but it has some unexpected behavior as it won't add category to current refinements and causes confusion for end users.
The second query is there to correctly filter the category and show it in current refinements widget.

Inwerpsel:
```javascript
        if (algoliaConfig.request.path.length > 0 && window.location.hash.indexOf('categories.level0') === -1) {
		    instantsearchOptions.searchParameters = {
                hierarchicalFacetsRefinements: {
                    'categories.level0': [algoliaConfig.request.path]
                }
            }
        }

```

There is probably a better way to check the url if any other category filters have been applied, but this applies the current category filter before the search is made.

Then, to make the category appear in the current refinements, you can just remove the lines that exclude this facet when on a category page:

```javascript
        $.each(algoliaConfig.facets, function (i, facet) {
            var name = facet.attribute;

            if (name === 'categories') {
                // if (algoliaConfig.isCategoryPage) {
                //    return;
                // }
                name = 'categories.level0';
            }

            if (name === 'price') {
                name = facet.attribute + algoliaConfig.priceKey
            }

            attributes.push({
                name: name,
                label: facet.label ? facet.label : facet.attribute
            });
        });```

Inwerpsel:
[...]

PR #852, opened 2017-06-16
Fixes the issue when back button returned all products instead of jus…

### LINKED ISSUE #825: Back button doesn't gets user back to the filtered category, but to all products instead
We have a "_Global Search on Category Page_" setting set to Yes and we have a problem.

if you go to any subcategory: shoes/styles/juniors (https://goo.gl/CFSl0a) , select any filter, like "size = 5" - you will see the filtered by size products. That's correct. 

After that try to hit Back button at browser and you will be back to shoes/styles/juniors page, but instead of seeing the products from that subcategory you will see the whole catalog, eventhough URL is correct...

I would expect, that we will see products from shoes/style/juniors - and now all the products.

JanPetr:
…t those on category page. At the same time removes the extra Algolia query on category page.

Fixes #825 #831

### DIFF
--- a/js/algoliasearch/instantsearch.js
+++ b/js/algoliasearch/instantsearch.js
@@ -59,6 +59,22 @@ document.addEventListener("DOMContentLoaded", function (event) {
 			}
 		};
 		
+		if (algoliaConfig.request.path.length > 0 && window.location.hash.indexOf('categories.level0') === -1) {
+			if (algoliaConfig.areCategoriesInFacets === false) {
+				instantsearchOptions.searchParameters = {
+					facetsRefinements: { }
+				};
+				
+				instantsearchOptions.searchParameters['facetsRefinements']['categories.level' + algoliaConfig.request.level] = [algoliaConfig.request.path];
+			} else {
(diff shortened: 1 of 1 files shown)


js/algoliasearch/instantsearch.js
```diff
@@ -59,6 +59,22 @@ document.addEventListener("DOMContentLoaded", function (event) {
 			}
 		};
 		
+		if (algoliaConfig.request.path.length > 0 && window.location.hash.indexOf('categories.level0') === -1) {
+			if (algoliaConfig.areCategoriesInFacets === false) {
```
Inwerpsel: Would be more readable to check for true:
``` javascript
if (algoliaConfig.areCategoriesInFacets) {
    instantsearchOptions.searchParameters = {
        hierarchicalFacetsRefinements: {
            'categories.level0': [algoliaConfig.request.path]
        }
    }
} else {
    ...
}
```

JanPetr: Makes sense @Inwerpsel,
thanks for the review. I changed it and will merge it when tests passes ...

PR #883, opened 2017-07-31
Show correct image for color specified by facet or query

JanPetr:
- [x] Add support in to autocomplete
- [x] Do the feature configurable
- [x] Fix configuration texts

irciwkzrdq

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
@@ -905,15 +905,16 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
             if ($attribute_resource) {
                 $attribute_resource->setStoreId($product->getStoreId());
 
+                $values = array();
+                $subProductImages = array();
+
                 /**
                  * if $value is missing or if the attribute is SKU,
                  * use values from child products.
                  */
                 if (($value === null || 'sku' == $attribute_name) && ($type == 'configurable' || $type == 'grouped' || $type == 'bundle')) {
-                    if ($value === null) {
-                        $values = array();
-                    } else {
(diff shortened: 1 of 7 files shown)


app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -905,15 +905,16 @@ public function getObject(Mage_Catalog_Model_Product $product)
...
                 /**
                  * if $value is missing or if the attribute is SKU,
                  * use values from child products.
                  */
                 if (($value === null || 'sku' == $attribute_name) && ($type == 'configurable' || $type == 'grouped' || $type == 'bundle')) {
-                    if ($value === null) {
-                        $values = array();
-                    } else {
-                        $values = array($this->getValueOrValueText($product, $attribute_name, $attribute_resource));
+                    if ($value !== null) {
```
julienbourdeau: I think this condition will always be true since we define `$values = array();` line 908.

js/algoliasearch/internals/frontend/common.js
```diff
@@ -20,24 +20,62 @@ document.addEventListener("DOMContentLoaded", function (e) {
...
 						return;
 					}
-
+					
 					colors.push(color.value);
+					
+					if (algoliaConfig.useAdaptiveImage === true) {
+						var re = /<em>(.*?)<\/em>/g;
+						var matchedWords = color.value.match(re).map(function (val) {
+							return val.replace(/<\/?em>/g, '');
```
julienbourdeau: Good idea 👍

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
JanPetr: On line 908 is set var `$values`, the condition checks var `$value` :)

Issue #877, opened 2017-07-24
If IS is enabled, don't load product from the backend


julienbourdeau:
By default, if you use instantsearch, the server will render the page with the product collection from Magento. This could be avoided since we're going to replace them with JavaScript.

https://discourse.algolia.com/t/dont-load-products-in-html/2056

JanPetr:
Will be part of the next release.

PR #886, opened 2017-08-04
Prevent backend rendering

JanPetr:
This feature prevents category pages to be rendered on backed when they use Algolia. As a result the rendering should be much faster. Implements #877 

**TODO:**

- [x] User-agent detection to run backend rendering for specified UAs
- [x] Description of the configuration input
- [x] Documentation page (https://github.com/algolia/magento/pull/73)
- [x] Test accessibility of website


alexgran:
Hi, works fine for me, even in heavily customised Apptha MArketplace based website.
I think we should use the same approach for product search?

JanPetr:
Thanks @alexgran!
I removed few more templates from backend rendering including search page.

Can you confirm that everything still works on your side and that you can see a performance improvement?

Thank you!

JanPetr:
Hello @alexgran,
did you have some time to try the latest changes in this PR?

alexgran:
Hi @JanPetr , currently on holiday, but got a developer on the task. Will have feedback next week.

alexgran:
@JanPetr Works like a charm. On a non optimised development server, this gives around 50% improvment on the page with all products.

JanPetr:
Thanks for your feedback @alexgran 💯

PR #903, opened 2017-08-31
special price fix

gregdev:
The special price attribute isn't selected when getting products; this fix ensures the special price is selected which allows it to be picked up in Algolia.


JanPetr:
Thank you for the contribution! :100: I create a new PR (#944) with only `special_price` fetch.

Thank you for brining this up!

PR #935, opened 2017-10-17
Use default for only_visible in product collection

julienbourdeau:
It seems that useTmpIndex is used to determine if we should select only visible product or not.


JanPetr:
Yes, and that's correct. When the extension indexes to TMP index, only visible products need to be processed. When it indexes directly to production index, all products need to be processed.
What is the reason / issue behind this PR?

julienbourdeau:
I was trying to figure out why some people get disabled product indexed and I couldn't understand the link between tmp index and only visible so I opened this PR "just in case".

Feel free to close it if it's incorrect.

JanPetr:
OK, let's close it for now. 
@julienbourdeau Could you send me a link to HS or thread with the issue?

PR #936, opened 2017-10-19
Allow translating hits widget empty template

mladenilic:
At the moment "No products" label cannot be translated.


JanPetr:
Hello @mladenilic, thank you for your contribution! 💯
There was a bug on instant search page, which didn't show the correct template. The no results page should be handled by [custom widget](https://github.com/algolia/algoliasearch-magento/blob/master/js/algoliasearch/instantsearch.js#L149-L189) and show translated "No products found" message and potentially suggested queries.
I made a fix in #942 - would that work for you? :)

I'm closing the PR now. Thank you very much for pointing the issue out! 👍

mladenilic:
Sure @JanPetr.

Thanks!

PR #956, opened 2017-11-07
Handle buggy onclick event on iOS

### LINKED ISSUE #902: weirdness on apple devices - have to click search results twice to go to a page
I'm running into an issue with the latest version 1.11 - I updated my dev site this morning. But this also happens on algolia's test site - https://magento.algolia.com/ - so I'll show you how to recreate it there.

Go to that test site on an iPhone or iPad. Search for something, like _pants_, and the autocomplete dropdown will show up. Try clicking on one of the results.

The first click with select the result (so you'll get the gray this-item-selected background), and the second click will actually go to the result.

And the problem is _only_ on apple devices - We've got iPhones (mostly 7s) and an 1st gen iPad, and it's the same behavior on all of them. It doesn't happen at all on our android devices (a wide variety of phone versions and tablets)

Also, I've got 1.8 in a production environment, and the issue doesn't happen there.

JanPetr:
Fixes #902 

@sam327, could you give it a try if it fixes your issue?

### DIFF
--- a/js/algoliasearch/autocomplete.js
+++ b/js/algoliasearch/autocomplete.js
@@ -105,5 +105,20 @@ document.addEventListener("DOMContentLoaded", function(event) {
 				fixAutocompleteCssSticky(menu);
 			});
 		});
+		
+		// Hack to handle buggy onclick event on iOS
+		$(algoliaConfig.autocomplete.selector).each(function () {
+			var data = $(this).data('aaAutocomplete');
+			var dropdown = data.dropdown;
+			var suggestionClass = '.' + dropdown.cssClasses.prefix + dropdown.cssClasses.suggestion;
+			
+			dropdown.$menu.on('touchstart', suggestionClass, function(e) {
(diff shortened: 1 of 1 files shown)


sam327:
Thank you! Put it into a dev site, and my initial testing looks good. Works as I'd expect on my older ipad. 

Throwing it at the rest of my office, as they've got a variety of apple devices, and we'll see how it goes.

JanPetr:
Thanks for coming back to me @sam327 💯 
Let me know if it's OK for your teammates as well and I'll merge :)

sam327:
So, this looks good to us. Tested on a half-dozen apple devices, and a couple non-apple ones (for good measure).

JanPetr:
Thanks for your help @sam327! 💯 It's been merged and will be part of the next release.

Haroenv:
This is amazing! Let's see how to incorporate this in autocomplete.js itself

PR #958, opened 2017-11-07
Don't prevent backend rendering on static content only categories

### LINKED ISSUE #954: "Prevent backend rendering" doesn't render static-block-only categories
There is an issue with preventing backend rendering.
It would prevent rendering for all category pages even if they are set to show a static block only.

If you are aware, in categories we have the option to show products, static block or both.
Seems like if that toggle is Yes, it does not render anything.

JanPetr:
Fixes #954

- [x] Add configuration to choose on what display mode should the backend rendering be enabled


royduin:
So this fixes the issue (just noticed it) where Algolia is rendered while `display_mode` on a category is set to `PAGE`? So Algolia should not load in that case.

JanPetr:
Hello @royduin! Actually this is just to render the static category page even though backend rendering is turned off.
But you're right, there is another issue which renders Algolia on static category pages as well. Thanks for bringing it up! I'll fix the issue as well 💯

royduin:
Cool, looking forward to the fix 😎

PR #964, opened 2017-11-09
Reindex Magento fulltext correctly for stores which has Algolia disabled

### LINKED ISSUE #890: catalogsearch_fulltext table isn't getting updated anymore?
We've got a shop with 3 stores, 2 of them are using Algolia and the 3th one is using the default Magento search. I noticed that the `catalogsearch_fulltext` table wasn't updated for a while (after the customer complained about new products not showing up in the search) so I've truncated the table, ran all indexes and noticed the table was still empty. When I disable Algolia completely and running the indexes the table gets filled again.

The shop is running on 1.8.0 with this fix applied: https://github.com/algolia/algoliasearch-magento/pull/742. I just updated to 1.10.0 just to test the index but the same issue.

JanPetr:
Fixes #890

@royduin, would you find some time to give it a try? I'd be really helpful! Thanks!


royduin:
Works perfectly! 😎 thank you!

JanPetr:
Thanks for the test, @royduin 💯 Merged and it'll be part of the next release.

Issue #979, opened 2017-11-28
Notice: Undefined offset: 0  in app/code/community/Algolia/Algoliasearch/Helper/Logger.php on line 34


ProxiBlue:
Saving a new CMS page in admin produces the error noted in subject of this ticket.
This is caused by storeid of 0 being used to return the name of the store for logging purposes. Store 0 (admin) is not in the given array.

![screenshot from 2017-11-28 14-57-30](https://user-images.githubusercontent.com/4994260/33306488-bde9418e-d44d-11e7-8582-013a62e57ef1.png)

JanPetr:
Will be part of the next release. Thanks @ProxiBlue!

PR #980, opened 2017-11-28
fix bug #979

ProxiBlue:


.gitignore
```diff
@@ -1,3 +1,5 @@
 .DS_Store
 vendor
 composer.lock
+.idea
```
JanPetr: Hello, can you remove this change from `.gitignore` file?
It's really specific to you environment and you can put it to your global `.gitignore` file.

Thanks!

PR #1025, opened 2018-05-07
Facet query rules

damcou:
Add facet query rules (adapted from M2 module)


HRR1337:
Maybe an stupid question, but what are Facet query rules ?

damcou:
Hi @HRR1337 ,

This feature allows the user to create automatically a query rule for a facet in the Instant Search Results Page section in the Magento back-office.

Here's some explanations about Algolia's query rules : https://www.algolia.com/products/query-rules .

Create a Query rule for facet when you want Algolia to automatically filter on a facet when a user searches for its value in query. It means when you create a Query rule for "color" facet and user is searching for query "red shorts", Algolia will internally filter results for color "red" and search only for query "shorts". It leads to more relevant searches. 

Hope this would help.

Best regards.

Damcou

PR #1028, opened 2018-05-22
Remove category from indexes of all store views

peresypkin:
**Summary**

If a store ID is not specified, that is $storeId == null, the category gets removed only from the index of the first store view in the list.

**Result**

The category gets removed from the indexes of all store views.


JanPetr:
Thank you, @peresypkin! 🙌

PR #1066, opened 2018-10-17
Fix autocomplete bottom links

damcou:
Remove "ors" category links when the categories facet is not active

### DIFF
--- a/js/algoliasearch/internals/frontend/common.js
+++ b/js/algoliasearch/internals/frontend/common.js
@@ -336,12 +336,21 @@ document.addEventListener("DOMContentLoaded", function (e) {
 
 					var ors = '';
 
-					if (keys.length > 0) {
-						ors += '<span><a href="' + keys[0].url + '">' + keys[0].key + '</a></span>';
+					var isCategoryFacetEnabled = false;
+					for (var i=0; i<algoliaConfig.facets.length; i++) {
+						if (algoliaConfig.facets[i].attribute == "categories") {
+							isCategoryFacetEnabled = true;
+						}
 					}
 
-					if (keys.length > 1) {
-						ors += ', <span><a href="' + keys[1].url + '">' + keys[1].key + '</a></span>';
+					if (isCategoryFacetEnabled) {
+						if (keys.length > 0) {
+							ors += '<span><a href="' + keys[0].url + '">' + keys[0].key + '</a></span>';
+						}
+
+						if (keys.length > 1) {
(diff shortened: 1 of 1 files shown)


js/algoliasearch/internals/frontend/common.js
```diff
@@ -336,12 +336,21 @@ document.addEventListener("DOMContentLoaded", function (e) {
...
+					for (var i=0; i<algoliaConfig.facets.length; i++) {
+						if (algoliaConfig.facets[i].attribute == "categories") {
+							isCategoryFacetEnabled = true;
+						}
 					}
 
-					if (keys.length > 1) {
-						ors += ', <span><a href="' + keys[1].url + '">' + keys[1].key + '</a></span>';
+					if (isCategoryFacetEnabled) {
+						if (keys.length > 0) {
```
JanPetr: Can those 2 conditions be replaced by a cycle?

js/algoliasearch/internals/frontend/common.js
```diff
@@ -336,12 +336,21 @@ document.addEventListener("DOMContentLoaded", function (e) {
 
 					var ors = '';
 
-					if (keys.length > 0) {
-						ors += '<span><a href="' + keys[0].url + '">' + keys[0].key + '</a></span>';
+					var isCategoryFacetEnabled = false;
```
JanPetr: Right now, the conditions happening no matter if there are some `keys` or not, which is not super effective. Can it be run only if some keys are present? Or can `keys` be retrieved only if categories are in facets? It's not effective to do both :)

js/algoliasearch/internals/frontend/common.js
```diff
@@ -337,11 +337,16 @@ document.addEventListener("DOMContentLoaded", function (e) {
...
 
 					if (keys.length > 0) {
-						ors += '<span><a href="' + keys[0].url + '">' + keys[0].key + '</a></span>';
-					}
-
-					if (keys.length > 1) {
-						ors += ', <span><a href="' + keys[1].url + '">' + keys[1].key + '</a></span>';
+						var orsTab = [];
+						for (var i = 0; i<algoliaConfig.facets.length; i++) {
+							if (algoliaConfig.facets[i].attribute == "categories") {
```
JanPetr: The `if` and `for` to check if categories are present are redundant to `for` on line 324. It's already checked that categories are present. If they wouldn't, `keys` would be an empty array.
So lines 341, 342, 347 (`break;`) can be removed.

At the same time, could you add spaces to `for` arguments?:
```
for (var i = 0; i < keys.length && i < 2 ; i++)
```

PR #1075, opened 2018-10-31
Exclude non-salable subproducts from price calculation

### LINKED ISSUE #1074: Wrong price range for configurable with out of stock subs
Fix configurable products' prices #978 includes sub products that are not salable (out of stock for example) in price range calculation.
This by itself is possibly a desired behavior but a further bug happens here when the subproducts are setup with special prices... for example we have a configurable with 2 sub products.
All are priced $100, with special price without end date $50.
When all the products are in stock the price calculation gives the $50 value. (which is correct)
When one of the subs goes out of stock the getFinalPrice call on him returns the full price value, and in the end the price calculation gives the "$50 - $100" value (which is incorrect).

I've opened a PR with a suggested fix.

Interghost:
Fixes #1074

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
@@ -600,15 +600,19 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
                     } else {
                         if (count($sub_products) > 0) {
                             foreach ($sub_products as $sub_product) {
+                                if (!$sub_product->isSalable()) {
+                                    continue;
+                                }
                                 $price = (double) $taxHelper->getPrice($product, $sub_product->getFinalPrice(), $with_tax,
                                                          null, null, null, $product->getStore(), null);
 
                                 $min = min($min, $price);
                                 $max = max($max, $price);
                             }
-                        } else {
+                        }
+                        if ($min > $max) {
                             $min = $max;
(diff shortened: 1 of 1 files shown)


app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -600,15 +600,19 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
                     } else {
                         if (count($sub_products) > 0) {
                             foreach ($sub_products as $sub_product) {
+                                if (!$sub_product->isSalable()) {
+                                    continue;
+                                }
                                 $price = (double) $taxHelper->getPrice($product, $sub_product->getFinalPrice(), $with_tax,
```
JanPetr: Could you add an empty line before `$price = ..." line?

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -600,15 +600,19 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
...
                                                          null, null, null, $product->getStore(), null);
 
                                 $min = min($min, $price);
                                 $max = max($max, $price);
                             }
-                        } else {
+                        }
+                        if ($min > $max) {
```
JanPetr: Could you add an empty line before `if (..." line?

Interghost:
Done

JanPetr:
You are awesome, thanks! 🚀

Issue #1095, opened 2019-01-09
Indexing Queue > 404 not found error


joell67:
Installed Algolia 1.15.0 in Magento 1.9.3.4 and 1.9.2.3.
Reindexed, cached, logged out etc. as usual.

Logged back in, navigated to System > Algolia Search > Configurations, entered all credentials in the fields, saved.

Then went to System > Algolia Search > Indexing Queue to reindex my catalog, but page is resulting in 404 not found.
Below is the path generated https://mysite.com/index.php/admin/algoliasearch_indexingqueue/index/key/abcdefghijkl123456789/ 

Same issue applies when navigating to Reindex SKU(s) - 404 error.
https://mysite.com/index.php/admin/algoliasearch_reindexsku/index/key/abcdefghijkl123456789/ .
Only Configurations page would load without issues.

To reproduce the issue, install the extension, navigate to either Indexing Queue or Reindex SKU(s) - there will be an error of 404 page not found.

Your help is appreciated.

joell67:
What do you mean by classes, where is that located?

I did rename the files to IndexingqueueController.php and ReindexskuController.php, also the block files were renamed in app/code/community/Algolia/Algoliasearch/Block/Adminhtml.

No more 404's! - Reindex SKU(s) is showing just fine, but Indexing Queue still shows a blank page in admin, nothing there. I must be missing something.

Thinking now maybe to change server settings to ignore capitalization, not sure what impact this will have globally though. I might be better off just renaming this extension files/folders.

Rgds,

jaystacey:
Algolia\Algoliasearch\Block\Adminhtml\IndexingQueue\Grid.php ~line 68

$this->addColumn('data', array(
            'header' => Mage::helper('algoliasearch')->__('Data'),
            'index' => 'data',
            'renderer' => 'Algolia_Algoliasearch_Block_Adminhtml_IndexingQueue_Grid_Renderer_Json'
        ));

to

$this->addColumn('data', array(
            'header' => Mage::helper('algoliasearch')->__('Data'),
            'index' => 'data',
            'renderer' => 'Algolia_Algoliasearch_Block_Adminhtml_Indexingqueue_Grid_Renderer_Json'
        ));

joell67:
This helped loading the page properly.
Thank you.

JanPetr:
Thank you for the report and the PR 🙌 I have just merged it and it will be released as a stable release soon.

tomekjordan:
[...]

PR #1096, opened 2019-01-10
Rename files to fix 404 issue #1095

jasuten:
The new indexing queue admin page gives a 404 on servers that are case sensitive. These changes rename the files and classes that are affected to fix them.

Fix for issue #1095


JanPetr:
[Thank you for the PR](http://bit.ly/2KHTYV3), @jasuten ❤️ It really helped us and we'll release it as a stable version soon.

PR #1100, opened 2019-01-16
Fix price_with_tax when customerGroups are enabled

rbrown:
When customerGroups are enabled, and magento is configured to display prices both **with** and **without** tax, the prices with tax were not being fetched with the search results.


JanPetr:
[Thank you for the PR](http://bit.ly/2KHTYV3), @rbrown! I'm happy to merge 🙌

PR #1117, opened 2019-04-05
Add max_record_size and indexing queue PID retry method 

bsuravech:
**Summary**
Reduce the number of "Record size is too big" errors by adjusting the record size according to plan through system admin and default the size to 10000. Update the retry method for stuck jobs with PIDs that have not be unassigned during system error (ie time out or max memory limits). 

**Result**
* Added max_record_size to pull limit from plan in system configuration 
* Changed the default max_record_size to 10000
* Added locked_at column to queue table and unlockStackedJobs() method to Queue model for jobs processing stuck after system error

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Model/Source/MaxRecordSize.php
+++ b/app/code/community/Algolia/Algoliasearch/Model/Source/MaxRecordSize.php
@@ -6,7 +6,6 @@ class Algolia_Algoliasearch_Model_Source_MaxRecordSize
 
     public function toOptionArray()
     {
-
         if (!$this->_options) {
             $options[] = Algolia_Algoliasearch_Helper_Config::DEFAULT_MAX_RECORD_SIZE;
             $proxyHelper = Mage::helper('algoliasearch/proxyHelper');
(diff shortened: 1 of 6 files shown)


app/code/community/Algolia/Algoliasearch/Model/Source/MaxRecordSize.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+class Algolia_Algoliasearch_Model_Source_MaxRecordSize
+{
+    protected $_options;
+
+    public function toOptionArray()
+    {
+        if (!$this->_options) {
+            $options[] = Algolia_Algoliasearch_Helper_Config::DEFAULT_MAX_RECORD_SIZE;
```
JanPetr: Does this mean that 10kb will be always the default value and will need to be changed manually by the customer?
If so - how about changing the form input to  "disabled" `text` and set it's value always based on what is coming from internal API? If internal API fails - use 10kb.

JanPetr: WDYT?

bsuravech: I rewrote the getMaxRecordSize() to fetch and save config if config does not have a value but cache the default value if for instance the API key isn't set. I remove the field as well since this is not changeable.

Issue #1043, opened 2018-07-02
Configurable prices are wrong after #978


rbrown:
Hello, Pull #978 doesn't 'fix' configurable pricing. I guess the previous behaviour was that configurable products didn't show a range of prices, but the code in #978 doesn't work correctly. 

#978 iterates through all the simple products attached to the configurable, and uses min/max on their prices, but that's not how configurable products are priced. 

Configurable products attribute values are assigned a modifier to the price (percent of absolute value) that is combined for each configurable attribute. You can see in `Mage_Catalog_Block_Product_View_Type_Configurable->getJsonConfig()` how the prices for each individual attribute value are sent to the frontend.

Because configurables are priced this way, it's entirely possible for the simple product pricing to not be the same as the price of the same product when presented as a child of a configurable product.

Please accept my apologies for not including a pull request to calculate the correct configurable price, but configurable pricing code always gives me a headache anytime I look at it.

damcou:
Hi @rbrown ,

Thanks for reaching us and for you interest on this issue.
I will try to have a look at it ASAP.
In the meanwhile, it could be super useful if you could give me some concrete examples of what the calculation is currently versus what it should be.

Thanks in advance.

Best regards.

rbrown:
Hi @damcou, I hope this image bomb is ok. In this example `Algolia_Algoliasearch_Helper_Entity_Producthelper->handlePrice` will set min and max to the same (£10) because that's the `price` of the configurable and associated products. But you can see that the configurable has price adjustments set on the `Associated Products` tab, and that the prices range from £10-£13.
<img width="887" alt="screen shot 2018-07-04 at 14 54 18" src="https://user-images.githubusercontent.com/233384/42281006-3db28036-7f9a-11e8-8a73-cd4eda977e50.png">
<img width="897" alt="screen shot 2018-07-04 at 14 20 54" src="https://user-images.githubusercontent.com/233384/42279569-1688ca3c-7f96-11e8-8a6e-4cf9e9d0acd3.png">
<img width="567" alt="screen shot 2018-07-04 at 14 18 13" src="https://user-images.githubusercontent.com/233384/42279534-f62f2470-7f95-11e8-833d-56cd9510b2fb.png">
<img width="573" alt="screen shot 2018-07-04 at 14 18 24" src="https://user-images.githubusercontent.com/233384/42279537-f7571f74-7f95-11e8-8de1-8f02e835423a.png">
[...]

PR #1119, opened 2019-04-09
Some fixes for customer group pricing

### LINKED ISSUE #1116: Wrong group price for configurable bundle products
**Do you want to request a *feature* or report a *bug*?**
Bug

**Bug: What is the current behavior?**
When I have a bundle product with configuration options (e.g. a helmet and goggles combined) the special price is used as group price although the defined group price is less than the special price. If no special price is defined the normal price is indexed and displayed instead of the group price.

More details to the product: The helmet has to be configured because the customer needs to select a size. For a goggle the customer can choose between a few different ones which also have a different price. So there is a price range e.g. from € 127.00 to € 151.00. The group price is defined as 10% discount.

**Bug: What is the expected behavior?**
The group price should be used instead of the special price or normal price.

**Bug: What is the proposed solution?**
    - 

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**
Magento 1.9.3
Algolia 1.15.0

rbrown:
Hello, I've noticed what I think are a couple of issues with customer group pricing, here's a pull request that fixes them.

I think there have been some inconsistencies with currencies/special prices etc. I don't think it will fix #1116 but I need to read through that again. I've essentially re-written the end of handlePrice in the Product Entity Helper, I'm going to try and comment on the specific changes to explain some of the things I did.

### DIFF
--- a/app/code/community/Algolia/Algoliasearch/Helper/Config.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Config.php
@@ -484,6 +484,8 @@ class Algolia_Algoliasearch_Helper_Config extends Mage_Core_Helper_Abstract
             foreach ($currencies as $currency) {
                 $attributes[] = $price.'.'.$currency.'.default';
                 $attributes[] = $price.'.'.$currency.'.default_formated';
+                $attributes[] = $price.'.'.$currency.'.default_original';
+                $attributes[] = $price.'.'.$currency.'.default_original_formated';
                 $attributes[] = $price.'.'.$currency.'.group_'.$groupId;
                 $attributes[] = $price.'.'.$currency.'.group_'.$groupId.'_formated';
                 $attributes[] = $price.'.'.$currency.'.group_'.$groupId.'_original_formated';
--- a/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
+++ b/app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
@@ -176,6 +176,7 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
             }
 
             $products = $products
+                ->addAttributeToSelect('special_price')
                 ->addAttributeToSelect('special_from_date')
                 ->addAttributeToSelect('special_to_date')
                 ->addAttributeToSelect('visibility')
@@ -496,14 +497,22 @@ class Algolia_Algoliasearch_Helper_Entity_Producthelper extends Algolia_Algolias
 
     protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_products, &$customData)
     {
+        /** @var Mage_Directory_Model_Currency $directoryCurrency */
+        $directoryCurrency = Mage::getModel('directory/currency');
+
+        /** @var Mage_Tax_Helper_Data $taxHelper */
+        $taxHelper = Mage::helper('tax');
+
+        /** @var Mage_Directory_Helper_Data $directoryHelper */
+        $directoryHelper = Mage::helper('directory');
+
         $fields = $this->getFields($product->getStore());
         $customer_groups_enabled = $this->config->isCustomerGroupsEnabled($product->getStoreId());
         $store = $product->getStore();
         $type = $this->config->getMappedProductType($product->getTypeId());
 
-        /** @var Mage_Directory_Model_Currency $directoryCurrency */
-        $directoryCurrency = Mage::getModel('directory/currency');
         $currencies = $directoryCurrency->getConfigAllowCurrencies();
+        $baseCurrencyCode = $store->getBaseCurrencyCode();
 
         if (Mage::helper('core')->isModuleEnabled('Mage_Weee') &&
             Mage::helper('weee')->getPriceDisplayType($product->getStore()) == 0) {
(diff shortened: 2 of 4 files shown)


app/code/community/Algolia/Algoliasearch/Helper/Config.php
```diff
@@ -484,6 +484,8 @@ public function getAttributesToRetrieve($groupId, $store)
             foreach ($currencies as $currency) {
                 $attributes[] = $price.'.'.$currency.'.default';
                 $attributes[] = $price.'.'.$currency.'.default_formated';
+                $attributes[] = $price.'.'.$currency.'.default_original';
+                $attributes[] = $price.'.'.$currency.'.default_original_formated';
```
rbrown: For consistency I added these in as the code appears to be expecting them

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -176,6 +176,7 @@ public function getProductCollectionQuery($storeId, $productIds = null, $only_vi
             }
 
             $products = $products
+                ->addAttributeToSelect('special_price')
```
rbrown: This was removed in ad4b976 but without it I couldn't see that the code was pulling in the special_prices for products anymore

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -512,102 +521,100 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
...
         foreach ($fields as $field => $with_tax) {
             $customData[$field] = array();
+            $field_price = (double) $taxHelper->getPrice($product, $product->getPrice() + $weeeTaxAmount, $with_tax, null, null, null, $store, null);
+            if ($canSpecialBeValid) {
+                $field_special_price = (double) $taxHelper->getPrice($product, $special_price, $with_tax, null, null, null, $store, null);
+            }
 
             foreach ($currencies as $currency_code) {
-                $customData[$field][$currency_code] = array();
+                $cdfc = array();
```
rbrown: I found it really hard to read  $customData[$field][$currency_code] everywhere, so I renamed this and added it at the end, I apologise if this breaks your code style

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -645,55 +652,59 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
                         $dashed_format = $this->formatPrice($min, false, $currency_code).' - '.$this->formatPrice($max,
                                 false, $currency_code);
 
-                        if (isset($customData[$field][$currency_code]['default_original_formated']) === false || $min <= $customData[$field][$currency_code]['default']) {
-                            $customData[$field][$currency_code]['default_formated'] = $dashed_format;
+                        if (isset($cdfc['default_original_formated']) === false || $min <= $cdfc['default']) {
+                            $cdfc['default_formated'] = $dashed_format;
 
                             //// Do not keep special price that is already taken into account in min max
-                            unset($customData['price']['special_from_date']);
```
rbrown: This code I'm not sure what it was trying to do, but I do not think there is a ['price'] key to $customData, because the prices under ['currency']

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -512,102 +521,100 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
...
                         $special_price,
                         false,
                         $currency_code
                     );
                 }
 
+                // configurable appears to assume that there are no customer_group_prices
+                // specified on the attached simple products, but needs re-writing anyway
+                // to correctly calculate the option price, not the simple price
                 if ($type == 'grouped' || $type == 'bundle' || $type == 'configurable') {
```
rbrown: From here, these changes are just to use the $cdfc variable, with the exception of the one comment I've added

app/design/frontend/base/default/template/algoliasearch/internals/configuration.phtml
```diff
@@ -35,7 +35,7 @@ $session = Mage::getSingleton('customer/session');
 $customerGroupId = $session->getCustomerGroupId();
 
 $priceKey = '.'.$currencyCode.'.default';
-if ($config->isCustomerGroupsEnabled($storeId) && $customerGroupId !== 0) {
```
rbrown: I believe this code was preventing transformHit in common.js from working correctly. 5505a50 changed the product and hit-item template to use group_0 rather than default, but because this file didn't change they were mis-matched

js/algoliasearch/internals/frontend/common.js
```diff
@@ -143,17 +142,33 @@ document.addEventListener("DOMContentLoaded", function (e) {
 				hit.price = hit.price[0];
 			}
 
-			if (hit['price'] !== undefined && price_key !== '.' + algoliaConfig.currencyCode + '.default' && hit['price'][algoliaConfig.currencyCode][price_key.substr(1) + '_formated'] !== hit['price'][algoliaConfig.currencyCode]['default_formated']) {
-				hit['price'][algoliaConfig.currencyCode][price_key.substr(1) + '_original_formated'] = hit['price'][algoliaConfig.currencyCode]['default_formated'];
+			if (Array.isArray(hit.price_with_tax)) {
```
rbrown: price_with_tax was being ignored if present, so I added it in

js/algoliasearch/internals/frontend/common.js
```diff
@@ -143,17 +142,33 @@ document.addEventListener("DOMContentLoaded", function (e) {
 				hit.price = hit.price[0];
 			}
 
-			if (hit['price'] !== undefined && price_key !== '.' + algoliaConfig.currencyCode + '.default' && hit['price'][algoliaConfig.currencyCode][price_key.substr(1) + '_formated'] !== hit['price'][algoliaConfig.currencyCode]['default_formated']) {
```
rbrown: So I could't quite get my head around what was going on with this if statement but all references from here onwards to price_key were broken because they expected price key to be .default or .group_0, when price key is .currency.key, so I removed the if statement

js/algoliasearch/internals/frontend/common.js
```diff
@@ -143,17 +142,33 @@ document.addEventListener("DOMContentLoaded", function (e) {
...
-				hit['price'][algoliaConfig.currencyCode][price_key.substr(1) + '_original_formated'] = hit['price'][algoliaConfig.currencyCode]['default_formated'];
+			if (Array.isArray(hit.price_with_tax)) {
+				hit.price_with_tax = hit.price_with_tax[0];
 			}
-			
-			if (hit['price'] !== undefined && hit['price'][algoliaConfig.currencyCode]['default_original_formated']
-				&& hit['price'][algoliaConfig.currencyCode]['special_to_date']) {
+
+			// price_key is in format .CURRENCY.KEY, fix that for array access
+			const price_key_as_key = price_key.replace('.' + algoliaConfig.currencyCode + '.', '');
```
rbrown: From here I re-wrote the final section, as the Product Entity Helper is now putting out 4 fields per group (group_x, x_formated, x_original, x_original_formated) I think I "simplified" this js by checking if the date is in the range for the special_price and if it is not then replacing x_formated with x_original_formated

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
```diff
@@ -512,102 +521,100 @@ protected function handlePrice(Mage_Catalog_Model_Product &$product, $sub_produc
...
+                $cdfc['default_formated'] = $this->formatPrice($price, false, $currency_code);
 
-                $special_price = (double) $taxHelper->getPrice($product, $product->getFinalPrice(), $with_tax, null, null, null, $product->getStore(), null);
-                $special_price = $directoryHelper->currencyConvert($special_price, $baseCurrencyCode, $currency_code);
-                $special_price += $weeeTaxAmount;
+                if ($canSpecialBeValid) {
+                    $special_price = $directoryHelper->currencyConvert($field_special_price + $weeeTaxAmount, $baseCurrencyCode, $currency_code);
 
+                    // this needs moving up a few levels
+                    $cdfc['special_from_date'] = Mage::app()->getLocale()->storeDate($store, $special_from_date)->getTimestamp();
```
rbrown: I'm using ->storeDate here because strtotime wasn't time zone aware, and configuration.phtml was using storeDate so I was get off by 3600 hours because of DST

rbrown:
I hope these changes are acceptable, understandable, the other change I didn't quite flag was that in the handlePrice I check if the special_price is valid or will be valid, if it's not we don't export any thing, otherwise we export the start/end date. I think the original JS assumed that special_prices were valid at the time of export, so a special_price that started in a few days would show up early in the search results.

rbrown:
Turns out I'm not the world's greatest tester. I'm not sure why I though changing a blank special expiry date to the current day would be a good idea when they day ticked over to the next day, but this latest commit  fixes that.

rbrown:
Re: configurable product prices - I've done some investigation, and I have a suggestion that would probably work.

When adding the prices to the original product collection, the min/max prices are being pulled out of the pricing index already, for group_0. 

For a product with price 13.75, group price 4.15 and one child set at +1.00 and one set at -1.00, The  index stores prices like this

|enity_id|customer_group_id|price|final_price|min_price|max_price|
|-|-|-|-|-|-|
|SKU|0|13.75|13.75|12.75|14.75|
|SKU|1|13.75|4.15|3.15|5.15|

for group_0/default prices the min/max are already loaded into the product object. For the later groups you could calculate the offsets (+1.00 -1.00) from group0 and the offsets to the groups.

Even if there is a special_price currently active you  can use the three prices (final/min/max) to calculate the correct offsets for the normal / special_price

I won't try and implement this until I hear your opinion on these other changes

damcou:
Hello @rbrown ,

First, apologies for the late answer and most of all, thanks a lot for taking the time to propose such big PRs !
I had a look at your changes, tried to understand everything, I must admit it's not that easy, especially when it's about this part of the extension which is really not easily maintainable  :( . 

The changes look good to me, but I'm struggling to find ways to test everything, in the description you mentioned "some inconsistencies with currencies/special prices" . Could you please tell me more about it ? 

Do you have any easy use-case that I can reproduce on my own, which will lead to those inconsistencies so I can test before/after the PR to see the differences? 

So far, I tried to activated the customer groups, setting different customer prices and a special price, and from what I see, the special price is taken into account. Should I choose a particular type of product ? Should I set a particular configuration ?

Thanks again for this.

rbrown:
Wow, yes, I see I never really said what it was fixing, sorry. Here's some product data
Screenshot 2019-05-16 at 13 22 26
Screenshot 2019-05-16 at 13 22 31
Screenshot 2019-05-16 at 13 22 39
Note the special price is in the future.

This is the dashboard with current develop branch
Screenshot 2019-05-16 at 13 17 59

This is the dashboard with my branch
Screenshot 2019-05-16 at 13 21 35

So, although the current branch knows there's special_from and special_to, it doesn't know what the special price is and won't be able to use it. In the new version I used `defaul`t et al for the special_price and `origina`l for the non special price, then in the js on the front end check if the current date is in the special period. I also include the `original` for all the groups, so that group 4 can say was £190 now £134, not was £200 because that's never the group4 price. _Maybe when there's no special price I should be exporting the default price as the group_original prices_

Other problems were the use of `product->getFinalPrice()` which is a cached call, and `product->getPriceModel()->getFinalPrice()` which is calculated and writes to final_price. So because the codes loops `with/without tax`, `currency` and `groups` the first time `product->getFinalPrice` is called in any loop after the very first, it returns the most recently called `product->getPriceModel()->getFinalPrice()` which will give you the price for the highest numbered group. In this example I set price to 200, and the group price for group 5 ( the highest number i have) to 190, current develop does this where I have one currency but use both with and without prices


Screenshot 2019-05-16 at 14 07 39
Screenshot 2019-05-16 at 14 07 44
This branch does this:
Screenshot 2019-05-16 at 14 04 46
Screenshot 2019-05-16 at 14 04 50

There was also some inconsistency between group_0 and default, some bits seemed to expect to replace group_0 with default and some didn't so `transformHit` wasn't always getting called. 

I hope that's a little clearer, sorry for not providing examples earlier

damcou:
Hello @rbrown ,

Thanks a lot for these examples, it's more clear to me now!
I tried to make the exact same tests on my own and I could reproduce everything 👍 

So if I sum up now, there are two main points : 

- **The special price timeframes are currently not handled by the extension**. If you set a special price in the future, this price is not indexed and won't be displayed once the timeframe is reached. 
Indeed, the changes you made seem to fix this. 
  - If the timeframe is reached, we display **[price]_original_formated** (strikethrough text) + **[price]_formated**
  - If the timeframe is not reached, we only display  **[price]_original_formated** 

You made a test with customer groups enabled, but I also tried the same test with customer groups disabled and it seems to fix it as well, does it make sense to you ?

- **There is a bad calculation of the prices when customer groups are enabled AND when we display excl. and incl. tax prices**. 
It is due to a bad use of the getFinalPrice() method that returns a cached value when we iterate more than once inside the "taxes fields" loop instead of redoing the calculation. 
To prevent that, you now use the getGroupPrice() method instead.

Did I miss something ?

For me the first point is an **improvment** of the handlePrice() method (it now handles special prices coming in a future timeframe), and the second point is a **fix** (price_with_tax object was just wrongly calculated) . Do you agree with that ?

If you confirm that I didn't miss or misunderstood anything, I'll ask one of my colleague to double check (as it's a quite sensitive part of the extension) and if it looks ok to him/her as well, I think we can merge it :) . 

Thanks again for this PR.

Have a nice week-end.
Best.

rbrown:
Hi, @damcou 

> * The special price timeframes are currently not handled by the extension. If you set a special price in the future, this price is not indexed and won't be displayed once the timeframe is reached.
> Indeed, the changes you made seem to fix this.
>   * If the timeframe is reached, we display [price]_original_formated (strikethrough text) + [price]_formated
>   * If the timeframe is not reached, we only display [price]original_formate
> You made a test with customer groups enabled, but I also tried the same test with customer groups disabled and it seems to fix it as well, does it make sense to you ?

Yes, it should be doing the right thing for `default_` prices, and then if customer groups are enabled, it should go through each one and do that properly too. Plus it should be consistent in how it generates this timestamp for the end/start times and the current time that is sent on each page load

> There is a bad calculation of the prices when customer groups are enabled AND when we display excl. and incl. tax prices.
> It is due to a bad use of the getFinalPrice() method that returns a cached value when we iterate more than once inside the "taxes fields" loop instead of redoing the calculation.
> To prevent that, you now use the getGroupPrice() method instead.

Yes, to what was happening. The fix is sort of the side effect of not using finalPrice. We can't use getFinalPrice, because it's affected by whether any special price is currently active. To handle future special prices correctly, and the group ones, we end up always using getPrice, getSpecialPrice and getGroupPrice

> If you confirm that I didn't miss or misunderstood anything

I think you've understood it correctly, thank you. There is one thing I must point out, excluding changes because I re-named variables, my changes all stop at the point where `handlePrice` tries to calculate the price of configurable products. Because the existing code is 'wrong' and I'm not sure how to fix it. 

We discussed this last year in #1043. I think, if you're happy with this pull request, that I can rewrite the configurable (and hopefully bundle/grouped we don't use them internally) to use the price index to get the correct offsets for the children prices and then calculate the correct prices in all groups. The problems that are obvious to me are:

* It relies on the price index for that product being correct
* It fixes the configurable pricing for vanilla magento installs, I am aware that there extensions that use the price of the simple children, and you may have customers using an extension like that, for whom the current behaviour is correct.

If we can't use the index, I can't see a solution that I'm happy with, short of copying wholesale the code magento already uses to calculate the configurable prices, and that makes my head hurt.

I hope you have a great weekend too

app/code/community/Algolia/Algoliasearch/Helper/Entity/Producthelper.php
damcou: You're right, this is painful to use in the rest of the function. On the other hand $cdfc isn't very meaningful and maybe too short. Could you please consider to change the variable name in something meaningful like $currencyData maybe ? Thx in advance :)

damcou:
Hi @rbrown ,

My colleague just had a second look at the PR and everything seems fine ! (except the name of the $cdfc variable as mentioned in the corresponding conversation.
Once it's changed, we will be able to merge it ! :) 

Regarding the conflgurable price calculation, yes I remember seeing your previous PR and it's still in our backlog to be processed in the future, but I can't guarantee any timeframe for it :( . 
But if you want to submit a new PR about it, feel free to do it, that would be very nice !

Best.
Damien

rbrown:
Hey @damcou The rename is done, I'll add the configurable prices to my backlog as well :p

PR #1129, opened 2019-04-26
Update search-insights and compatibility 

bsuravech:
Updated search-insights.js version and updated Click Analytics to make it compatible with new changes. This will allow for Personalization tracking with the updated version.


js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,20 +1,31 @@
...
 	// "Click" on instant search page
 	$(document).on('click', algoliaConfig.ccAnalytics.ISSelector, function() {
 		var $this = $(this);
-		trackClick($this.data('objectid'), $this.data('position'));
+		var lastResults = analyticsHelper.getLastResults();
+
+		// want to track results returned
+		if (lastResults) {
+			trackClick(lastResults.index, $this.data('objectid'), $this.data('position'), lastResults.queryID ? lastResults.queryID : lastResults._rawResults[0].queryID);
+		}
```
tkrugg: This might not work for infinite hits. Worth checking whether it's a use case we want to cover.

js/algoliasearch/internals/frontend/search-insights.js
```diff
@@ -1 +1 @@
...
...e provide `objectIDs` to be reported");if(!e.objectIDs)throw new Error("required objectIDs parameter was not sent, conversion event can not be properly sent without");this.sendEvent("conversion",e)}.bind(this),this.convertedFilters=function(e){if(!e)throw new Error("No params were sent to convertedFilters function, please provide `filters` to be reported");if(!e.filters)throw new Error("required filters parameter was not sent, conversion event can not be properly sent without");this.sendEvent("conversion",e)}.bind(this),this.viewedObjectIDs=function(e){if(!e)throw new Error("No params were sent to viewedObjectIDs function, please provide `objectIDs` to be reported");if(!e.objectIDs)throw new Error("required objectIDs parameter was not sent, view event can not be properly sent without");this.sendEvent("view",e)}.bind(this),this.viewedFilters=function(e){if(!e)throw new Error("No params were sent to viewedFilters function, please provide `filters` to be reported");if(!e.filters)throw new Error("required filters parameter was not sent, view event can not be properly sent without");this.sendEvent("view",e)}.bind(this),this.setUserToken(this.ANONYMOUS_USER_TOKEN),this.processQueue()}});
```
tkrugg: there has been a new version of insights since then it might be worth updating it :)
There was no breaking changes.

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -43,47 +54,48 @@ algoliaBundle.$(function ($) {
...
+	AlgoliaAnalytics.clickedObjectIDsAfterSearch(clickData);
 }
 
-function trackConversion(objectID) {
-	AlgoliaAnalytics.conversion({
-		objectID: objectID.toString()
+function trackConversion(indexName, objectIDs) {
+	AlgoliaAnalytics.convertedObjectIDsAfterSearch({
+		index: indexName,
+		eventName: "Conversion",
```
tkrugg: if this is for perso, we should pick a better Event Name.
@tvweinstock any recommandations or patterns that emerged for perso event names?

js/algoliasearch/click_conversion_analytics.js
tvweinstock: @tkrugg do HBK is only using pagination, not infinite hits I believe. @bsuravech do you know if the M1 module uses infinite hits at all?

js/algoliasearch/click_conversion_analytics.js
tvweinstock: @tkrugg is this for perso? I thought that `convertedObjectIDsAfterSearch ` was for clickAnalytics and `convertedObjectIDs` was for personalisation.

js/algoliasearch/click_conversion_analytics.js
bsuravech: @tvweinstock We do offer infinite hits as apart of the configuration:  

**Algolia Search > Instant Search Results Page > Enable Infinite Scroll** 

If we want to incorporate it into the extension, we should cover infinite hits.

js/algoliasearch/internals/frontend/search-insights.js
bsuravech: I've updated this to v1.3.0 :) Thanks for letting us know!

js/algoliasearch/click_conversion_analytics.js
bsuravech: @tkrugg @tvweinstock  this PR was specifically made to update the search-insights.js and click_conversion_analytics.js for the insights update. We're not adding personalisation in as it's more involved with fetching customerIDs and what not.

js/algoliasearch/internals/frontend/search-insights.js
tkrugg: it's perfect!

js/algoliasearch/click_conversion_analytics.js
tkrugg: alright, sorry I got a little confused because I thought we were doing this specifically for allowing perso. Because if I read this correctly, click analytics already work with the old implem. Am I forgetting something?

tkrugg: @bsuravech so I talked about this with @tvweinstock and apparently we need to support perso as well. Is this something we'd want to tackle in another PR ?

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -43,47 +54,48 @@ algoliaBundle.$(function ($) {
...
-	}
-
-	AlgoliaAnalytics.click(clickData);
+	AlgoliaAnalytics.clickedObjectIDsAfterSearch(clickData);
 }
 
-function trackConversion(objectID) {
-	AlgoliaAnalytics.conversion({
-		objectID: objectID.toString()
+function trackConversion(indexName, objectIDs) {
```
tkrugg: we should add the queryID here.

bsuravech:
@damcou When you have a moment, can you review the order conversion part of this item?

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
-		&& algoliaConfig.ccAnalytics.orderedProductIds.length > 0) {
-		$.each(algoliaConfig.ccAnalytics.orderedProductIds, function (i, objectId) {
-			trackConversion(objectId);
-		});
+	if (algoliaConfig.ccAnalytics.conversionAnalyticsMode === 'place_order') {
+
+		if (typeof algoliaOrderConversionJson !== 'undefined') {
+			$.each(algoliaOrderConversionJson, function(idx, itemData) {
+				if (itemData !== null && typeof itemData.objectID !== 'undefined') {
+					trackConversion(itemData.indexName, itemData.objectID, itemData.queryID);
```
tkrugg: amazing!

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
 	// "Click" in autocomplete
 	$(algoliaConfig.autocomplete.selector).each(function () {
-		$(this).on('autocomplete:selected', function (e, suggestion) {
-			trackClick(suggestion.objectID, suggestion.__position, suggestion.__queryID);
+		$(this).on('autocomplete:selected', function (e, suggestion, dataset, context) {
+
+			var sources = analyticsHelper.sources;
+			var source = sources.find(function(e) {
+				return e.name == dataset;
+			});
```
tkrugg: 1. does this still work if I use the twice the same dataset?
2. [find](https://caniuse.com/#feat=array-find) is not supported on IE11, it's best to use the _.find or use [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
-		applicationID: algoliaConfig.applicationId,
+		appId: algoliaConfig.applicationId,
 		apiKey: algoliaConfig.instant.apiKey
 	});
 
 	// "Click" in autocomplete
 	$(algoliaConfig.autocomplete.selector).each(function () {
-		$(this).on('autocomplete:selected', function (e, suggestion) {
-			trackClick(suggestion.objectID, suggestion.__position, suggestion.__queryID);
+		$(this).on('autocomplete:selected', function (e, suggestion, dataset, context) {
```
tkrugg: context is not used.

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
+
+			var sources = analyticsHelper.sources;
+			var source = sources.find(function(e) {
+				return e.name == dataset;
```
tkrugg: I'm not sure about this line.
dataset is an integer, so it should be something like `sources[dataset -1]`

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
+			});
+
+			trackClick(source.indexName, suggestion.objectID, suggestion.__position, suggestion.__queryID);
```
tkrugg: source is the result of `find`, so it can be undefined. It's best to check before trying to access .indexName

js/algoliasearch/click_conversion_analytics.js
```diff
@@ -1,89 +1,108 @@
...
-	if (algoliaConfig.ccAnalytics.conversionAnalyticsMode === 'place_order'
-		&& algoliaConfig.ccAnalytics.orderedProductIds.length > 0) {
-		$.each(algoliaConfig.ccAnalytics.orderedProductIds, function (i, objectId) {
-			trackConversion(objectId);
-		});
+	if (algoliaConfig.ccAnalytics.conversionAnalyticsMode === 'place_order') {
+
+		if (typeof algoliaOrderConversionJson !== 'undefined') {
+			$.each(algoliaOrderConversionJson, function(idx, itemData) {
+				if (itemData !== null && typeof itemData.objectID !== 'undefined') {
```
tkrugg: you can just write this as `if (itemData && itemData.objectID)`

app/code/community/Algolia/Algoliasearch/Model/Observer/Conversion.php
```diff
@@ -0,0 +1,63 @@
...
+    public function setAlgoliaParamsToSession(Varien_Event_Observer $observer)
+    {
+        /** @var Mage_Core_Controller_Front_Action $controllerAction */
+        $controllerAction = $observer->getEvent()->getControllerAction();
+        $params = $controllerAction->getRequest()->getParams();
+
+        $checkoutSession = Mage::getSingleton('checkout/session');
+        if (!$this->_isOrderConversionTrackingEnabled($checkoutSession->getQuote()->getStoreId())) {
+            return;
+        }
```
damcou: Could this condition be moved on top of the method ?

js/algoliasearch/click_conversion_analytics.js
bsuravech: I've switched it to use `filter` instead which should return an array. We can check if filter returns a count before accessing the source.

Issue #1140, opened 2019-11-06
Can not save magento CMS page


alex-magebinary:
When save the CMS page, one of the function count is expecting an array, but string is given.
For example:
The wrong one:
`Algolia_Algoliasearch_Helper_Entity_Pagehelper->getPages('1', '10')`

The right one:
`Algolia_Algoliasearch_Helper_Entity_Pagehelper->getPages('1', array('10'))`

The error:
`2019-11-06T03:41:04+00:00 DEBUG (7): Exception message: Warning: count(): Parameter must be an array or an object that implements Countable  in /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Helper/Entity/Pagehelper.php on line 32
Trace: #0 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Helper/Entity/Pagehelper.php(32): mageCoreErrorHandler(2, 'count(): Parame...', '/var/www/html/....', 32, Array)
#1 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Helper/Data.php(203): Algolia_Algoliasearch_Helper_Entity_Pagehelper->getPages('1', '10')
#2 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Model/Observer.php(163): Algolia_Algoliasearch_Helper_Data->rebuildStorePageIndex('1', '10')
#3 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Model/Resource/Engine.php(49): Algolia_Algoliasearch_Model_Observer->rebuildPageIndex(Object(Varien_Object))
#4 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Model/Resource/Engine.php(105): Algolia_Algoliasearch_Model_Resource_Engine->addToQueue('algoliasearch/o...', 'rebuildPageInde...', Array, 1)
#5 /var/www/html/.modman/algoliasearch/app/code/community/Algolia/Algoliasearch/Model/Observer.php(126): Algolia_Algoliasearch_Model_Resource_Engine->rebuildPages('1', '10')
#6 /var/www/html/app/code/core/Mage/Core/Model/App.php(1374): Algolia_Algoliasearch_Model_Observer->savePage(Object(Varien_Event_Observer))
[...]

PR #1141, opened 2019-11-08
Update CMS Page Indexing

bsuravech:
**Summary**
Correct issue regarding \Algolia_Algoliasearch_Model_Observer::savePage() to pass in array instead of string for type check.

Related Issues: #1140 

**Result**
Update the CMS Page save event to pull from updated changes and add optimisation by indexing data from collection instead of cms page model load.


alex-magebinary:
Hi How far away for this to be merge to master?

PR #1149, opened 2020-01-15
fix: Insights URL for CC Analytics only

bsuravech:
**Summary**
Insights URL was not optional and includes the URL even if cc analytics was not enabled.

**Result**
Removed unused event. Added `enabled` as a ccAnalytics parameter and added condition to the makeURL function.

### DIFF
--- a/js/algoliasearch/instantsearch.js
+++ b/js/algoliasearch/instantsearch.js
@@ -2,6 +2,10 @@ document.addEventListener("DOMContentLoaded", function (event) {
 	algoliaBundle.$(function ($) {
 
 		function makeUrlForInsights(baseUrl, objectID, queryID, indexName) {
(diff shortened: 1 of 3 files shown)


js/algoliasearch/instantsearch.js
```diff
@@ -2,6 +2,10 @@ document.addEventListener("DOMContentLoaded", function (event) {
 	algoliaBundle.$(function ($) {
 
 		function makeUrlForInsights(baseUrl, objectID, queryID, indexName) {
+			if (algoliaConfig.ccAnalytics.enabled !== true) {
```
tkrugg: can this be simplified to `algoliaConfig.ccAnalytics.enabled` ? unless it's not always a boolean

PR #1161, opened 2020-05-19
Avoid rules removal issues when doing full reindex

maxiloc:


bsuravech:
I wonder if there is a benefit here to upgrade the PHP Client to the latest v 1.28. It is not good practice to modify the library directly. We should looked at the `scopedCopyIndex()` method introduced in 1.24 of the PHP client. 

https://github.com/algolia/algoliasearch-client-php/releases/tag/1.24.0

We'll discuss this internally but leaving a comment here for visibility on product decision.

@damcou

bsuravech:
@maxiloc Without updating the client, performing a reindex with the changes in this PR throws an exception error that was returned by the API:

```
Algolia Search Products index process unknown error:
AlgoliaSearch\AlgoliaException: The source and the destination are the same in /var/www/htdocs/.modman/algoliasearch-magento/lib/AlgoliaSearch/Client.php:995
Stack trace:
#0 /var/www/htdocs/.modman/algoliasearch-magento/lib/AlgoliaSearch/Client.php(812): AlgoliaSearch\Client->doRequest(Object(AlgoliaSearch\ClientContext), 'POST', '*******.algo...', '/1/indexes/mage...', Array, Array, 1, 30)
#1 /var/www/htdocs/.modman/algoliasearch-magento/lib/AlgoliaSearch/Client.php(366): AlgoliaSearch\Client->request(Object(AlgoliaSearch\ClientContext), 'POST', '/1/indexes/mage...', Array, Array, Array, 1, 30)
#2 /var/www/htdocs/.modman/algoliasearch-magento/app/code/community/Algolia/Algoliasearch/Helper/Algoliahelper.php(212): AlgoliaSearch\Client->copyIndex('magento_193_def...', 'magento_193_def...', Array)
```

maxiloc:
I've updated the PR. I removed an `elseif` that was needed. I added it back, should do the trick

bsuravech:
@maxiloc It looks good now, rules and synonyms seems to be copying over. 

I'm not going to merge this PR as I would like to update the PHP Client. You can create a diff patch for your customer.

maxiloc:
👍 , I linked this PR changes, so if can just close it (and not delete it) that would be good as it would stay accessible.

Thanks

bsuravech:
@maxiloc Do you think there would be any consequences of having search core removing the empty parameter returned when we search for rules? This would be an immediate fix compared to our patching and both extensions would benefit from it.

maxiloc:
No that would work but I also don't think it makes sense to fix that in the engine as from the engine pov the behavior is correct and removing things from the rule when it's valid could be confusing as well.

Also It's also way better to copy rules/synonyms with one call, it will prevent more issues in the future.

bsuravech:
@maxiloc I've been trying to ascertain if there are consequences or missing actions needed for the replicas as it seems our copy methods in the helper add forwardToReplicas on saveRules()/saveSynonyms(). Anything we should consider here?

maxiloc:
I would say synonyms needs to be copied on the replica while rules do not but then it really depends on the usecase.

bsuravech:
@maxiloc Thanks for your investigative work on this! We've corrected this in PR #1166. Closing this PR out!

